PR #2, opened 2018-04-11
Use environment variables for example's settings

expobrain:
Settings in the `examples` directory are very specific but we want to allow the users to use their settings without changing the code.

### DIFF
--- a/examples/_config.py
+++ b/examples/_config.py
@@ -1,4 +1,6 @@
-dsn = 'exasol-dev.mlan:8563'
-user = 'SYS'
-password = 'exasol'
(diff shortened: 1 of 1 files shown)


wildraid:
Good point as well!

Fixed in: https://github.com/badoo/pyexasol/commit/2dd19536a268bc5f0f01ae24fb6a09daa77a1834#diff-5ea60a2266d552845cacc062bd3de40b

I've used "default" option instead of "or".
I've also used standard Exasol names for parameters (EXAHOST, EXAUID, EXAPWD, EXASCHEMA).

But the main idea remains the same.

Thank you!

PR #4, opened 2018-04-16
Fixed parsing datetime's string

expobrain:
Question: why checking if microseconds is bigger than 20?

### DIFF
--- a/pyexasol/mapper.py
+++ b/pyexasol/mapper.py
@@ -32,6 +32,6 @@ def exasol_mapper(val, data_type):
     elif data_type['type'] == 'TIMESTAMP':
         return datetime.datetime(int(val[0:4]), int(val[5:7]), int(val[8:10]),       # year, month, day
                                  int(val[11:13]), int(val[14:16]), int(val[17:19]),  # hour, minute, second
-                                 int(val[20:24]) * 1000 if len(val) > 20 else 0)     # millisecond (if available)
+                                 int(val[20:24]) if len(val) > 20 else 0)     # microseconds (if available)
     else:
         return val


wildraid:
It means that TIMESTAMP format is longer than 20 characters.
Example: **2000-01-01 00:00:00.** (mind the dot)

This is exactly 20 characters. If value is longer, it means that fractional part was included, and we try to parse it.

Timestamp precision of Exasol is milliseconds, not microseconds. To check:
`sql
ALTER SESSION SET NLS_TIMESTAMP_FORMAT='YYYY-MM-DD HH24:MI:SS.FF6';
SELECT TIMESTAMP'2018-01-01 03:04:05.123456';
`
Result: 2018-01-01 03:04:05.123000

There are no microseconds in Exasol, and FF6 format is not default anymore. Current code supports three reasonable timestamp formats:

- YYYY-MM-DD HH24:MI:SS.FF0 (no fractional part);
- YYYY-MM-DD HH24:MI:SS.FF3 (3 decimal places, milliseconds);
- YYYY-MM-DD HH24:MI:SS.FF6 (6 decimal places, fake microseconds, legacy compatibility);

If we accept suggested change, it will break both FF3 and FF6 formats.

wildraid:
But I agree it may need some additional commenting and support for FF1 and FF2 formats. I'll add it.

wildraid:
Simplified code a bit, added support for fractional part of any length:
https://github.com/badoo/pyexasol/commit/8d7b9c3d24fe29edad9ec9b5b4ff172ea0a0605e

Looks like ljust() is implemented in C. Performance should be ok.

PR #15, opened 2019-02-20
Default host

expobrain:
Use a sensible name for the default host name


wildraid:
Ok, it's a good point. Let's merge it.

PR #16, opened 2019-02-27
Fix concurrency issue in ExaConnection req method

tkilias:
We found a bug with concurrent queries against the websocket api. If you use a ExaConnection in two concurrent threads to execute queries, we receive similar error messages from the database as the following one:

`Received packet type not expected while command execution. (Session: XXXXXXXX)`

The websocket api doesn't support concurrent requests for a connection. This change adds a lock around the req method which ensures a sequential execution of the queries. Furthermore, we added tests which reproduce the bug when the lock is not used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ build/
 
 # Visual Studio Code
 .vscode
+
+venv
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,3 +68,4 @@ script:
     - python examples/23_snapshot_transactions.py
   # - python examples/24_script_output.py  disabled due to connectivity problems
     - python examples/25_overload.py
+    - python tests/test_concurrent_queries.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,3 +69,4 @@ script:
   # - python examples/24_script_output.py  disabled due to connectivity problems
     - python examples/25_overload.py
     - python tests/test_concurrent_queries.py
+    - python tests/test_threadsafety_peformance.py
--- a/pyexasol/connection.py
+++ b/pyexasol/connection.py
@@ -12,7 +12,8 @@ from . import callback as cb
 from . import constant
 from . import utils
 
-from .exceptions import ExaQueryError, ExaQueryTimeoutError, ExaRequestError, ExaCommunicationError, ExaRuntimeError
+from .exceptions import ExaQueryError, ExaQueryTimeoutError, ExaRequestError, ExaCommunicationError, ExaRuntimeError, \
+    ExaConcurrencyError
 from .statement import ExaStatement
 from .logger import ExaLogger
 from .formatter import ExaFormatter
@@ -80,7 +81,6 @@ class ExaConnection(object):
         :param client_name: Custom name of client application displayed in Exasol sessions tables (Default: PyEXASOL)
         :param client_version: Custom version of client application (Default: pyexasol.__version__)
         """
-        self.request_lock = threading.Lock() # to prevent concurrent requests against the database
 
         self.dsn = dsn
         self.user = user
@@ -127,6 +127,10 @@ class ExaConnection(object):
         self.ws_req_time = 0
 
         self._init_logger()
+
+        self._request_lock = threading.RLock() # to prevent other threads to use this connection
+        self.acquire() # pin the the connection the the creating thread
+
         self._init_format()
         self._init_ws()
         self._init_json()
@@ -136,6 +140,23 @@ class ExaConnection(object):
         self.get_attr()
 
 
+    def acquire(self):
+        self.logger.debug("try to acquire %s", threading.get_ident())
+        if not self._request_lock.acquire(blocking=False):
+            raise ExaConcurrencyError(self, "A connection can be used only by the owning process.")
+        else:
+            self.logger.debug("acquired %s", threading.get_ident())
+            self.owning_thread = threading.get_ident()
+
+    def release(self):
+        self.logger.debug("try to release %s", threading.get_ident())
+        if not self._request_lock.acquire(blocking=False):
+            raise ExaConcurrencyError(self, "Only the owning process can release the connection.")
+        else:
+            self._request_lock.release()
+            self.owning_thread = None
+            self.logger.debug("released %s", threading.get_ident())
+
 
     def execute(self, query, query_params=None) -> ExaStatement:
         """
@@ -143,8 +164,14 @@ class ExaConnection(object):
         Return ExaStatement object
         """
         self.last_stmt = self.cls_statement(self, query, query_params)
+        return self.last_stmt
+
+
+    def _execute_without_lock(self, query, query_params=None) -> ExaStatement:
+        self.last_stmt = self.cls_statement(self, query, query_params, req_method=self._req)
 
         return self.last_stmt
+
 
     def execute_udf_output(self, query, query_params=None):
         """
@@ -218,6 +245,7 @@ class ExaConnection(object):
         return self.import_from_callback(cb.import_from_pandas, src, table, callback_params)
 
     def export_to_callback(self, callback, dst, query_or_table, query_params=None, callback_params=None, export_params=None):
+        self.acquire()
         from .http_transport import ExaSQLExportThread, ExaHTTPProcess, HTTP_EXPORT
 
         if not callable(callback):
@@ -240,7 +268,6 @@ class ExaConnection(object):
         try:
             http_proc = ExaHTTPProcess(self.ws_host, self.ws_port, compression, self.encryption, HTTP_EXPORT)
             http_proc.start()
-
             sql_thread = ExaSQLExportThread(self, http_proc.get_proxy(), compression, query_or_table, export_params)
             sql_thread.set_http_proc(http_proc)
             sql_thread.start()
@@ -250,7 +277,6 @@ class ExaConnection(object):
 
             http_proc.join()
             sql_thread.join()
-
             return result
         except Exception as e:
             # Close HTTP Server if it is still running
@@ -264,6 +290,7 @@ class ExaConnection(object):
             raise e
 
     def import_from_callback(self, callback, src, table, callback_params=None, import_params=None):
+        self.acquire()
         from .http_transport import ExaSQLImportThread, ExaHTTPProcess, HTTP_IMPORT
 
         if callback_params is None:
@@ -312,6 +339,7 @@ class ExaConnection(object):
         Get proxy strings from each child process
         Pass proxy strings to parent process and use it for export_parallel() call
         """
+        self.acquire()
         from .http_transport import ExaSQLExportThread
 
         if export_params is None:
@@ -336,6 +364,7 @@ class ExaConnection(object):
         Get proxy strings from each child process
         Pass proxy strings to parent process and use it for import_parallel() call
         """
+        self.acquire()
         from .http_transport import ExaSQLImportThread
 
         if import_params is None:
@@ -406,48 +435,57 @@ class ExaConnection(object):
         return [{'host': ip_address, 'port': self.ws_port, 'idx': idx} for idx, ip_address
                 in enumerate(itertools.islice(itertools.cycle(ret['responseData']['nodes']), pool_size), start=1)]
 
+
     def req(self, req):
-        with self.request_lock:
-            """ Run WebSocket request synchronously """
-            self.ws_req_count += 1
 
-            # Build request
-            send_data = self._json_encode(req)
-            self.logger.debug_json(f'WebSocket request #{self.ws_req_count}', req)
+        """ Run WebSocket request synchronously """
+        self.acquire()
+        return self._req(req)
 
-            start_ts = time.time()
+    def _req(self, req):
 
-            # Send request, wait for response
-            try:
-                self._ws_send(send_data)
-                recv_data = self._ws_recv()
-            except websocket.WebSocketException as e:
-                raise ExaCommunicationError(self, str(e))
+        """ Run WebSocket request synchronously """
+        self.ws_req_count += 1
 
-            self.ws_req_time = time.time() - start_ts
+        # Build request
+        send_data = self._json_encode(req)
+        self.logger.debug_json(f'WebSocket request #{self.ws_req_count}', req)
 
-            # Parse response
-            ret = self._json_decode(recv_data)
-            self.logger.debug_json(f'WebSocket response #{self.ws_req_count}', ret)
+        start_ts = time.time()
 
-            # Updated attributes may be returned from any request
-            if 'attributes' in ret:
-                self.attr = {**self.attr, **ret['attributes']}
+        # Send request, wait for response
+        try:
+            self._ws_send(send_data)
+            recv_data = self._ws_recv()
+        except websocket.WebSocketException as e:
+            raise ExaCommunicationError(self, str(e))
 
-            if ret['status'] == 'ok':
-                return ret
+        self.ws_req_time = time.time() - start_ts
 
-            if ret['status'] == 'error':
-                # Special treatment for "execute" command to prevent very long tracebacks in most common cases
-                if req.get('command') == 'execute':
-                    if ret['exception']['sqlCode'] == 'R0001':
-                        cls_err = ExaQueryTimeoutError
-                    else:
-                        cls_err = ExaQueryError
+        # Parse response
+        ret = self._json_decode(recv_data)
+        self.logger.debug_json(f'WebSocket response #{self.ws_req_count}', ret)
 
-                    raise cls_err(self, req['sqlText'], ret['exception']['sqlCode'], ret['exception']['text'])
+        # Updated attributes may be returned from any request
+        if 'attributes' in ret:
+            self.attr = {**self.attr, **ret['attributes']}
+
+        if ret['status'] == 'ok':
+            return ret
+
+        if ret['status'] == 'error':
+            # Special treatment for "execute" command to prevent very long tracebacks in most common cases
+            if req.get('command') == 'execute':
+                if ret['exception']['sqlCode'] == 'R0001':
+                    cls_err = ExaQueryTimeoutError
                 else:
-                    raise ExaRequestError(self, ret['exception']['sqlCode'], ret['exception']['text'])
+                    cls_err = ExaQueryError
+
+                raise cls_err(self, req['sqlText'], ret['exception']['sqlCode'], ret['exception']['text'])
+            else:
+                raise ExaRequestError(self, ret['exception']['sqlCode'], ret['exception']['text'])
+
+
 
     def _login(self):
         ret = self.req({
@@ -519,6 +557,7 @@ class ExaConnection(object):
         options = {
             'timeout': self.socket_timeout,
             'skip_utf8_validation': True,
+            'enable_multithread': True
         }
 
         if self.encryption:
--- a/pyexasol/connection.py
+++ b/pyexasol/connection.py
@@ -45,6 +45,7 @@ class ExaConnection(object):
             , lower_ident=False
             , json_lib='json'
             , verbose_error=True
+            , threadsafety=0
             , debug=False
             , debug_logdir=None
             , udf_output_host=None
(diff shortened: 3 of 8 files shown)


wildraid:
Ok, multi-threading is an interesting topic. Let's discuss it.

1) Did you try to set option `enable_multithread` for websocket client? It can be passed to `websocket.create_connection()` function as extra parameter.

2) It may create many other unwanted effect, I am afraid. For example, please consider following code:

```
C.export_to_pandas(...)
st = C.last_statement()
```

What if Python switches threads between 1st line and 2nd line, and another thread executes some other query? I guess, we will have an inconsistent result of "last_statement()".

Similar problems may also apply to other functions relying on specific state. What if someone tries to fetch from one dataset in multiple threads? Etc.

I think it might be the best practice to declare database driver as not thread-safe and prevent all the hassle related to it. 

If someone needs parallelism for basic queries, new connection might be opened and closed in each thread. If someone needs parallelism for massive data reads or writes, parallel HTTP transport is a good choice not limited by single CPU and GIL-related problems.

Please let me know what you think. And thank you for testing this and raising this topic in general. :)

wildraid:
Also, I can potential major issue with transaction context.

For example, let's imagine we have two threads and connection with autocommit=False.

```
Thread 1:
C.execute("DELETE FROM table WHERE ...")
C.execute("INSERT INTO table ... ");
C.commit()

Thread 2:
C.execute("DELETE FROM table WHERE ...")
C.execute("INSERT INTO table ... ");
C.commit()
```

It is possible to encounter thread switch between DELETE and INSERT, which may lead to unwanted COMMIT between those two statements, which will break data consistency. So transactions and multi-threading cannot work together very well, if I understand it right.

Same applies to session variables and context variables. What if someone changes `query_timeout` or runs `ALTER SESSION SCRIPT_LANGUAGE='...';` in another thread? It's very fragile even with locks, I am afraid.

tkilias:
No, I didn't try enable_multithread, but it is absolutly a valid point. I didn't know about it. I give it a try. 

To your second comment. Your correct if you use the method last_statement it can get nasty. Actually, didn't thought about it, because it is not in my typical usage pattern. Your right it is actually better to create one connection per thread and the concurrency with threads is in python limited by the GIL.  For multi-processing you need either way multiple connection. I agree with you for real parallism the http transport is better suited . To declare it as not thread-safe is probably the better long-term solution, but I think currently it doesn't hurt to have the lock in place. It will prevent the most obvious bugs. 

By the way, what happen when you use mutliprocessing and share the connection. Did some tried this. I can assume, if you are not careful with multiprocessing it gets shared between the processes, or is it not serializable.

wildraid:
Websocket-client uses basic `socket.socket()` internally with no extra synchronisation, so `multiprocessing` will break it for sure.

By the way, current HTTP transport implementation uses both `threading` and `multiproessing`. We create a separate process first, and we add an extra thread only after that. This thread is running in controlled environment.

If we add an extra threading before calling HTTP transport wrapper function, it will lead to all threads being "forked" into separate process.

Complex stuff indeed :)

tkilias:
One question to the enable_multithread, I can't find any documentation for it. Where did you read/heard about it? 

And I think, the problem here is the database, which assumes a single threaded use of the websocket. It is probably the only correct way, anyway, because query parallism is done in the database by mutliple transactions. Your examples show very well the pitfalls of mutli-threading with one connection.

The only scenario where mutliple queries from different threads would be safe, is a read-only scenario with only executes or export_to_*, but the multiple threads wouldn't accelerate the queries in anyway.

wildraid:
`enable_multithread` is documented in the source code of library. If you have a good Python IDE, you may Ctrl + click on `create_connection()` call.

I see it quite often, especially for complex cases and rarely used options.

tkilias:
So, basically for http transport we also have to ensure, that we don't somehow produce a deadlock or something like that. phew. To declare pyexasol as not thread safe sounds very attractive, but as I said for the time being the lock shouldn't hurt, but I think I have to add a test for the http transport as well. The good thing, the http transport examples seems to work fine, leaves the question what happens if we extend it with multiple threads which use the connection.

tkilias:
Ok, I am officially blind. Found it, thx. And no, it does'nt solve the issue. Probably, because the exasol websockets api with request - response pairs works and the lock of `enable_multithread` only the request or the reponse guards, but not the whole pair.

tkilias:
One last idea for today. I mentioned transactions in a previous comment. They are probably the solution to the problem. The unpredictable reorderung of statements can happen, because we lock at the lowest level. What we actually need, is locking at the level of groups of statements which are actually transaction. However, for the locking in pyexasol, we don't need real database transactions, we only need a way to group statements and acquire before the group a lock and release it after the group. This more or less describe the function of context-manager in python. Here an example, how the thing could look like:

`
def func1(c):
	do_some_work()
	with c.begin_transaction() as t:
		t.execute(...)
		t.last_Statement()

def func2(c):
	with c.begin_transaction() as t:
		t.execute(...)
		t.execute(...)
	do_some_work()

c=pyexasol.connect(...)
t1 = Thread(target=func1, args=(c))
t1.start
t2 = Thread(target=func2, args=(c))
t2.start()
t1.join()
t2.join()
`

We also can add a shortcut for use case where it is known that the order of the statements doesn't matter, in that case the call to c.execute(...) acquires the lock executes the statement and release the lock directly.

Further, we can optionally couple the the begin_transaction context-manager with real database transactions.

wildraid:
Hmm, what if I just add `with` context support for connection objects? So each thread will look like:

```
with pyexasol.connect(...) as C:
    C.execute('DELETE FROM table')
    C.execute('INSERT INTO table ... ')
    C.commit()
```

We'll have full parallelism, more clarity and better performance overall.

tkilias:
Yes, sure also a possibility. I am actually, not sure if it would solves the problem. For example, someone hast the idea, ok than lets start the threads within the with-statement. I have to think about this case a bit more,

And a few other questions which come to my mind: 

- How expensive is the creation of connection? 
- Ccan we make backward compatible. Especially, the case where existing code already uses the connect() without context-manager and the order of the queries is not of interest. I would assume we protect than each query separately and print a warning that this can lead to out of order queries with multi- threading.

wildraid:
Opening connection is relatively cheap compared to typical execution time of Exasol queries. It normally takes 20-50ms.

However, unlike classic DBMS like MySQL, Exasol cannot handle too many parallel queries and heavy spikes. If some information calculated in Exasol should be accessed frequently, it might be a good idea to add queue and store results into MySQL. And put this MySQL under actual load from users.

wildraid:
I did some extra research on how thread safety is handled by other DB drivers.

For example, please take a look at MySQL driver for Python: https://mysqlclient.readthedocs.io/user_guide.html#functions-and-attributes

> 
> **threadsafety**
> 
> Integer constant stating the level of thread safety the interface supports. This is set to 1, which means: Threads may share the module.
> 
> The MySQL protocol can not handle multiple threads using the same connection at once. Some earlier versions of MySQLdb utilized locking to achieve a threadsafety of 2. While this is not terribly hard to accomplish using the standard Cursor class (which uses mysql_store_result()), it is complicated by SSCursor (which uses mysql_use_result(); with the latter you must ensure all the rows have been read before another query can be executed. It is further complicated by the addition of transactions, since transactions start when a cursor execute a query, but end when COMMIT or ROLLBACK is executed by the Connection object. Two threads simply cannot share a connection while a transaction is in progress, in addition to not being able to share it during query execution. This excessively complicated the code to the point where it just isn’t worth it.
> 
> The general upshot of this is: Don’t share connections between threads. It’s really not worth your effort or mine, and in the end, will probably hurt performance, since the MySQL server runs a separate thread for each connection. You can certainly do things like cache connections in a pool, and give those connections to one thread at a time. If you let two threads use a connection simultaneously, the MySQL client library will probably upchuck and die. You have been warned.

If MySQL avoids messing with threads for connection objects, why should we? MySQL is capable of handling many queries per second, and it has tools for connection pooling (connection reset).

Exasol can handle very limited parallelism and has no tools to reset connection to it's original state. So potential benefits from pooling and connection re-usage is even more limited compared to MySQL.

tkilias:
In my experience, parallel select queries are ok, depending on the complexity of the query. Write heavy workloads are limited by the disk performance, but this is normal. 

To execute multiple queries in parallel, we need multiple connections anyway. In case we doesn't want to fix the problem, we should either document it, or if possible detect a shared usage of a connection and throw a warning or exception.

Besides this, you said that the connection creation takes only 20-50ms. This is quite fast and it doesn't hurt to much to create new connections. So we could provide a connection factory which gets the connection information and creates new connections with a context-manager. The context-manager holds an RLock which can be acquired by the same thread multiple times, but other threads can not acquired it when it is already acquired by another thread. Each query execution needs to acquired the lock and only gets executed if the same thread that created the connection requests the query execution. If another thread tries to execute a query with a connection which doesn't belong to it, the query execution should throw an exception with a clear error message, that connection can only be used by the creating thread. 
If we don't want the factory, we also can use the existing connect(...) method which creates a connection with a context-manager and a RLock.

Here an example for [RLock](https://gist.github.com/tkilias/dd4e03ba11ce52361197b852b6d7ebf5) and here how the the factory could work:

    connection_factory=pyexasol.create_factory(...)
    with connection_factory.connect() as C:
        C.execute('DELETE FROM table')
        C.execute('INSERT INTO table ... ')
        C.commit()

wildraid:
I'll do some more research on weekend and come back to you. We may discuss it over the phone next week, no problem :)

I just have one general question for you before I start. Do you know any specific use cases when threading and connection factory / pooling is truly beneficial for Exasol?

Only two things come to my mind:

1. Web-application connecting directly to Exasol to read small amount of aggregated data;
2. Unit and functional tests;

If you know other cases, please let me know.

tkilias:
Sorry, I had no time to answer before the weekend. I can imagine the following additonal use cases for concurrent queries:

- realtime reporting
- realtime visualization
- generating batch for training of ML models from datasets to large for a full export
- streaming

However, the connection factory was only an idea to easer create new connections without having to transfer the credentials to each thread. Pooling could be necessary if there is a connection limit or the 20-50ms for creating a new connection is to much.

To solve the actual issue with concurrent requests, I pushed a new commit which uses a RLock instead of a Lock, such that only the owning process can use the ExeConnection. In the beginning the process which creates the ExaConnection is the owning process and as long the user code doesn't call the release-method this doesn't change. If the user code calls the release-method, another thread can acquire the ownership by either call the acquire-method or executing a request. The httptransport thread is a special case. Because this thread is controlled by the owning thread, the httptransport thread can circumvent the lock without releasing it. If a thread tries to execute a request which is not the owning thread the ExaConnection throws a ExaConcurrencyError.

wildraid:
Ok, thank you for pushing this topic forward.

I did some investigations during weekend. My current thoughts:


### 1. Existing PyODBC and TurbODBC libraries have `threadsafety=1`

PyODBC: https://github.com/mkleehammer/pyodbc/wiki/Module#threadsafety
TurbODBC: https://github.com/blue-yonder/turbodbc/blob/35f8725d758aa28b763c48571bc9b1fd1ebaf6c8/python/turbodbc/api_constants.py#L6

It means that pretty much all existing code is affected by similar issues while trying to share connections.

### 2. Thread locking adds some overhead for single-threaded code

SQLite3 module actually has protection mechanism similar to your suggestion. However, it is implemented in C, and there is an option to disable it.

https://github.com/python/cpython/blob/842acaab1376c5c84fd5966bb6070e289880e1ca/Modules/_sqlite/connection.c#L1120-L1133

For pure Python implementation such lock introduces extra overhead of maintaining `RLock` object, which makes single-threaded code a bit slower. Also, the package code becomes more complex and harder to understand.

Overhead is small, but it will affect all existing code in a negative way.

### 3. It might give false impression that threading is good idea

But it is probably not a good idea due to performance reasons.

No matter how many threads we run, they will be blocked either by GIL (single CPU core bottleneck) or Exasol connection executing queries (I/O bottleneck). In theory, we might get some performance boost with 2 threads, when one thread is waiting results from Exasol and another thread is doing something CPU intensive. But any serious parallelism (10-20-30-100 threads) will not be possible due to nature of Python.

---

I think the best course of action at this moment is to:

1. Add `threadsafety=1` to `ExaConnection` object, add comments and documentation entry.
2. Discourage simultaneous usage of connection objects by multiple threads. Non-simultaneous usage is perfectly fine.
3. Encourage usage of `subprocess` and `multiprocessing` to achieve true parallelism. Just fork processes and create individual connections normally. Operating system and Exasol will do the rest in the most optimal way. This approach might be scaled even on multiple servers, if one server is not enough.

I hope this explanation makes sense. Thanks for your time and effort (regardless of outcome).

tkilias:
Yes, locks introduce a bit of overhead and mutli-threading with python in general is not a good idea. Trust me, I already had to parallelize python code and I used in the end some form of multiprocessing or C extensions. 
However, I am not sure how large the overhead for the single lock exactly is, but I agree it is a good idea to enable or disable it via a flag. If I understand the links correctly, we are at the right track with only allowing one thread to use a connection. 
My suggestion would be, that I add the flag to the current implementation with the RLock and add comments to the affected methods. Addtionally, I can add a test which measures the overhead of the lock. 
Furthermore, I think it is a good idea to document the threading behavior in the Readme as well and also to document the error message which can happen if you use pyexasol with multi-threading without activating the flag.

tkilias:
Ok, one thing before I add the flag, what is the default. From your comment, I would assume not thread-safe is the default.

wildraid:
Why don't we just follow PyODBC / TurbODBC approach and allow people to break their code with sub-optimal usage of threading, so they can naturally start digging and go forward with multiprocessing or parallel HTTP transport?

If we actually allow "safe" multi-threading and add test for it, people will start copy-pasting this code without realising why performance is so bad in the end.

Also, please note how you had to introduce a hack to allow HTTP transport to bypass the lock. People might use threading for some technical reasons, and this is fine and it does not require any locks. Connection can be created in one thread, but used in another thread safely.

I can see at least one "hackish" case in future: command `abort`. It is not implemented yet, but if we ever do it, it should be called from separate thread.

I feel like merging tread locks will introduce more problems than it solves.

wildraid:
In order to improve error message, I could catch error code which is specific for `Received packet type not expected while command execution` and show more detailed exception about lack of multi-threading support.

tkilias:
Actually, with the RLock we don't allow multi-threading, because only one thread at the time is allowed to use the connection, if the threadsafety is enabled. 

The problem with only catching the error code and returning a meaningful exception without a lock, is that this error only occurs randomly. You can use a shared connection with mutliple threads and for pure luck don't encounter the error, but have unpredictable order of queries. I suggest, I try shortly how large the overhead of the lock and the flag for the lock is and than we can decide, if we use the lock or only catch the exception. And, if we should use the flag, we should in case threadsafety is disabled, catch the error anyway and throw a more descriptive error.

Yes, it is true the solution for httptransport is but "hackish", but in that case it is safe to circumvent the lock, because the import/export_from_callback methods are guarded by the lock and the import/export thread is under tied control of the owning thread of the connection.

tkilias:
I added the flag and implemented a performance test, here the results on my laptop with exasol in a virtual machine as docker-db.

wildraid:
Good evening,

It might take some time for me to figure out and test potential edge cases. What is currently on my mind:

1) Potential deadlock problems due to simultaneous usage of `multithreading` and `multiprocessing`: https://bugs.python.org/issue6721. It's a curious little ticket which is almost 10 years old. Interesting reading.

2) Sometimes we may have to lock not one, but multiple requests. For example, all `set_attr()` requests should be followed by sequential `get_attr()` due to minor inconsistencies of Exasol WebSocket protocol. Also, some functions in `ext.py` are doing multiple `_execute()` calls.

3) Transactions are still a major concern. Do we need extra protection for them or not?

wildraid:
I am still not fully convinced about viability of "officially supporting" threadsafety=2.

Please note that there are some other features which exist in Exasol protocol, but are not supported by PyEXASOL on purpose. Some examples:

1. Prepared statements. Dynamic parameters cause performance problems, are not supported by UDF's and cannot be used for database identifiers.

2. Sub-connections. Parallel HTTP transport is much better for Python in high-parallel scenarios, and it also offers more flexibility. For example, you cannot open more sub-connections than number of Exasol nodes, but you can make any number of HTTP connections.

3. INSERT statements with VALUES. It is highly recommended to use IMPORT instead in all cases, even if amount of data is low. It guarantees that sensitive data will never leak into AUDIT SQL logs by accident, and it also makes investigations much easier.

The main point is to not simply allow people do certain things, but to softly enforce some best practices.

tkilias:
I see, there are a lot more things to consider. I will think about them. To your last comment. That was actually, my intention with RLock, because it gives you a clear error message, that you can't use a connection with multiple threads. If you catch the error from the DB and return a clearer error message, you still have the randomness of this error message.
Currently, I can't they anything intelligent to the potential issue with multiprocessing, but I can assume that you are right. My question is, could we test it?
Concerning the setattr and getattr, is there are enclosing method or is this open to the user. If we have a enclosing method we could protect this method, too. If remember correctly, I did this already for some methods. And either way, the connection is only owned by one thread. The user have to release it willingly and if we think release is overall no good idea, we also can remove it and say, a connection can only be used in thread where it was created. 
I will think tommorow further about your comments and come back if I have some ideas. However, thanks for the really good discussion.

wildraid:
@tkilias , what do you think about following approach?

1. Guard `req()` method with simple `threading.Lock`.
2. If another thread tries to send any command (except `abort` command) - throw concurrency exception asking people to rethink their threading practices.
3. Do not add any extra complications.

Reasons:
- It will allow smart people to implement controlled multi-threading if they need it. They won't have to think about extra locks inside Python library.
- Bad practice of accessing same connection simultaneously from multiple threads will be automatically discovered and discouraged. The chance of such application to go through testing and Q&A phase without getting concurrency exception is very low.
- `Abort ` command will be available naturally and without an extra effort.
- Code will remain clear, with no extra locking \ unlocking for each method and no general locking.

wildraid:
I've checked C implementation of `threading.Lock`. It seems to be fine. Depending on OS, they try to use condition variable first, and mutex is the second choice.

I think, it should be fine to add it to `req()` method even for single-threaded scenarios. Communication and network overhead is MUCH bigger than this type of locking.

tkilias:
Short question: Do you mean Lock or RLock, because their semantic is very different, but I assume you mean Lock, because of the part with the Q&A phase. A Lock would allow multi-threaded usage except in the case of simultaneous access. If we use a Lock with blocking=false, we could detect this case, but it will contain randomness and requires very good tests to detect it reliable. So, basically it gets detected reliable by profifient developer, because they will write tests if they use it conccurrency, but less profifient developers could use the library accidently multi-threaded, produce unexpected results and the error will caught only by accident. Not sure, if it is good approach, but to be fair I don't know, what are the chances for the second scenario. If we use this approach, we have to state clearly in the documentation that mutlti-threading is bad practice, if you do it, you should test it very well.

wildraid:
@tkilias , please check the proposed update. It is not pushed to PyPi yet and can be changed if necessary: https://github.com/badoo/pyexasol/commit/8357a85c11300f1f5cc84fe9df5c9a2879b74808

Main points to consider:

- Choosing `Lock` vs. `RLock` looks like a better option for me. Technically, it is perfectly fine to offload query execution into separate thread. In some cases it might be the only way to go. Please see an example for [`abort_query()`](https://github.com/badoo/pyexasol/blob/master/examples/27_abort_query.py) call;

- `RLock` may break backwards compatibility for companies currently using threading in a proper way (e.g. Revolut). They'll have to update their code.

- `RLock` won't help beginner developers much. Most of Python standard functions are not thread-safe, including `print()` and `.write()` to file. There is no safety net. Understanding of synchronisation primitives is crucial;

- I've checked internal implementation of `threading.Lock` on Linux and MacOS. It looks really good. There are fast code paths for cases when (1) there are no other active threads or (2) lock is in unlocked state. Overhead is incredibly low for our situation.

tkilias:
Ok, the abort_query examples makes me clear what you meant. You are right this gets really ugly with RLock. Ok, if we already have companies which use the current way of multi-threading in proper way, then it is not a good idea to break this. To conclude this, with the Lock we do what we can to help the user without breaking something and additionally we state in the documentation that no two threads should use the connection at the same time. You convinced me that this is good way to proceed. Again, thanks for the great discussion and for fixing the bug.

wildraid:
Sweet! I've uploaded 0.5.4 to PyPi. Let's see how it goes.

Thank you for raising this problem in the first place. I was thinking about it during original implementation, but decided to spend more time on other things (like HTTP transport).

PR #20, opened 2019-03-26
CI: Remove sudo require

foxyblue:
Travis doesn't require running sudo to build and run the tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 __pycache__/
 *.py[cod]
+*.mypy_cach/
+
 
 dist/
 build/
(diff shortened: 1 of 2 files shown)


wildraid:
OK, it looks really simple and harmless. Let's merge it.

I think there are many things which can be improved in automated testing.

1) Windows support;
2) `sleep 120` potentially can be replaced with waiting script of some sort;

foxyblue:
@wildraid You might want to add those two points to Issues, that way one can be aware of how to contribute. 😄 

The Help Wanted label also makes an issue discoverable - but it all depends on what makes it easier for you.

PR #21, opened 2019-03-27
TST: Add tests to utils

foxyblue:
I have added a test to the method that transforms the `dsn`.
I wanted to capture the current behaviour before changing it to accept strings:
>`"127.0.0.1:8888,127.0.0.2:8889"`

I have also added a `pytest` dependency for the "extra_requires: `dev`".

I think the tests add clarity as to the transformations of the `dsn` string.

Let me know what you think :)


wildraid:
I'll double-check current behaviour in ODBC / JDBC drivers and come back to you in the next 1-3 days. Exasol drivers might parse such DSN's, but do they capable of utilising different ports?

I'll also think about potential complications in HTTP transport related to this change. Just in case.

Thank you.

foxyblue:
>but do they capable of utilising different ports?

^ This I am not aware of.

Just to be clear I haven't allowed the dsn to receive strings of this nature: `"127.0.0.1:8888,127.0.0.2:8889"`. I have only tested the current implementation, in this PR.

Thanks,

wildraid:
@foxyblue , please also note an interesting edge case regarding to host ranges.

For example, we have 5 nodes cluster + 1 reserve node.

Default "Connection string" recommended by ExaOperation: `myexasol1..5:8563`. However, if one data node fails and reserve node goes online, recommended "Connection string" will be `myexasol1..3,myexasol5..6:8563`. Please note that port is applied to both ranges in such case.

wildraid:
@foxyblue , I'll push some code to add support of multi-port DSN's when I get back from holidays. I have the patch ready, but it's a sensitive part, which might break on complex edge cases.

I will also research pytest best practices. I see it's approach is different from *Unit, so I have to learn a few things about it first.

foxyblue:
I am happy to rewrite tests in a way you are familiar with

wildraid:
@foxyblue , I still have some important stuff to do this week. But next week I am planning to push bigger update and refactor connection object. This will include multi-port DSN's.

As far as I know, Revolut found a temporary solution at this moment, so it should not be very urgent. Is it correct?

foxyblue:
@wildraid This is correct, thanks ^_^

wildraid:
@foxyblue , please check the latest version `0.6.0`. It should work with multi-port DSN connection strings.

I did not make my mind about pytest yet, but I've added an example which can be executed as normal Python script: https://github.com/badoo/pyexasol/blob/master/examples/29_dsn_parsing.py

It can be used for experimentation.

Thank you.

foxyblue:
This is great!
Thanks, I'll close this PR.

PR #31, opened 2019-08-22
Set autocommit & query_timeout from attributes

inytar:
At the moment the `ExaConnection.autocommit` &
`ExaConnection.query_timeout` can desync from what is set on the Exasol
server, this can lead to confusion. e.g.

```python
import pyexasol
import _config as config

C = pyexasol.connect(config.dsn, user.config.user, password=config.password,
                     autocommit=False)
C.set_autocommit(False)
assert C.attr['autocommit'] is False  # Works
assert C.autocommit is False  # Fails.
```

Here this inconsistency is fixed by setting the `autocommit` &
`query_timeout` from the attributes every time we get them.

### DIFF
--- a/pyexasol/connection.py
+++ b/pyexasol/connection.py
@@ -399,6 +399,8 @@ class ExaConnection(object):
         })
 
         self.attr = ret['attributes']
+        self.autocommit = self.attr['autocommit']
+        self.query_timeout = self.attr['query_timeout']
 
     def set_attr(self, new_attr):
         self.req({


wildraid:
Well, those properties are just temporary variables to hold connection parameters. They are not mean to be accessed or modified, but Python does not really allow to have "private" properties.

I'll consider moving all connection parameters into separate dictionary and clean up the list of properties a little bit to reduce confusion.

wildraid:
@inytar , I've pushed some changes to reduce confusion. Please check the changelog: https://github.com/badoo/pyexasol/blob/master/CHANGELOG.md

The right way to check autocommit state is: `C.attr['autocommit']`. Connection option `C.autocommit` is no longer available as raw property, but it can now be accessed by `C.options['autocommit']`.

Hope it helps!

inytar:
Thanks for looking into this!

PR #32, opened 2019-08-25
Feature/add option to return rows as named tuples

jabadia:
With this change, users can optionally request that rows are returned as namedtuples, with column names being the namedtuple field names

### DIFF
--- a/pyexasol/connection.py
+++ b/pyexasol/connection.py
@@ -53,6 +53,7 @@ class ExaConnection(object):
             , encryption=False
             , fetch_dict=False
             , fetch_mapper=None
+            , named_tuples=False
             , fetch_size_bytes=constant.DEFAULT_FETCH_SIZE_BYTES
             , lower_ident=False
             , quote_ident=False
@@ -81,6 +82,7 @@ class ExaConnection(object):
         :param encryption: Use SSL to encrypt client-server communications for WebSocket and HTTP transport (Default: False)
         :param fetch_dict: Fetch result rows as dicts instead of tuples (Default: False)
         :param fetch_mapper: Use custom mapper function to convert Exasol values into Python objects during fetching (Default: None)
+        :param named_tuples: return rows as namedtuples with column names instead of plain tuples (Default: False)
         :param fetch_size_bytes: Maximum size of data message for single fetch request in bytes (Default: 5Mb)
         :param lower_ident: Automatically lowercase identifiers (table names, column names, etc.) returned from relevant functions (Default: False)
         :param quote_ident: Add double quotes and escape identifiers passed to relevant functions (export_*, import_*, ext.*, etc.) (Default: False)
@@ -111,6 +113,7 @@ class ExaConnection(object):
 
             'fetch_dict': fetch_dict,
             'fetch_mapper': fetch_mapper,
+            'named_tuples': named_tuples,
             'fetch_size_bytes': fetch_size_bytes,
             'lower_ident': lower_ident,
             'quote_ident': quote_ident,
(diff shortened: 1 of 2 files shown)


wildraid:
@jabadia, ok, it's an interesting idea. I did a quick research about `namedtuple`.

It appears that creation of `namedtuple` is significantly slower than creation of both `tuple` and `dict`. Access time by index is almost identical to `tuple`, but access time by column name is much slower compared to `dict`.

You may find some tests on recent Python versions here: https://gist.github.com/zed/cc530d8d8c2f149aa6670b68c97c0f9c#gistcomment-2583111

Since Exasol is meant to process large volumes of data, I am not sure if `namedtuple` can be recommended from performance standpoint. Also, mixed access both by index and name might add some confusion to the code.

In my opinion, normal tuples and dicts are better choices both in terms of performance and code clarity.

However, I see there is a new feature available in Python 3.7 called "data classes". Maybe it can be used efficiently instead of named tuples.

jabadia:
Thanks for your answer and research.

> It appears that creation of namedtuple is significantly slower than creation of both tuple and dict. Access time by index is almost identical to tuple, but access time by column name is much slower compared to dict.

I agree that `namedtuples` introduce a performance overhead. That's why it should be optional.

> In my opinion, normal tuples and dicts are better choices both in terms of performance and code clarity.

The reason behind this change is to make rows behave similarly to rows returned by `pyodbc` (https://github.com/mkleehammer/pyodbc/wiki/Row).

wildraid:
I did some more research on this. I can share some interesting findings:

1) `Row` object in PyODBC is not a named tuple. They use different structure instead, which is mutable.

2) Some database drivers actually implemented `NamedTupleCursor`. For example, this is true for PostgreSQL database driver. However, please see some [limitations mentioned in code]( https://github.com/psycopg/psycopg2/blob/be8e1a26326edb6b094a764797460181ae8b33b6/lib/extras.py#L376). There are problems with column names starting with underscore or number. Also, they have to maintain LRU cache for efficiency.

3) Named tuples can only have maximum 255 arguments in Python 3.6. It [was fixed](https://bugs.python.org/issue18896) in Python 3.7, but it won't be possible to read wide dataset in Python 3.6.

4) Performance implications are dire. It's just a massive waste of CPU cycles when processing large amounts of data.

I don't see any real use case when having named tuples can be beneficial for quality of the code.

However, I've extended [25_overload](https://github.com/badoo/pyexasol/commit/99183cb2588bfb89872684f2d3e05f8cba82e6d3) example to show how you can still use named tuples in your code if you wish to do so.

```
class NamedTupleExaStatement(pyexasol.ExaStatement):
    def __next__(self):
        row = super().__next__()
        return self.cls_row(*row)

    def _init_result_set(self, res):
        super()._init_result_set(res)
        self.cls_row = collections.namedtuple('Row', self.col_names)


class NamedTupleExaConnection(pyexasol.ExaConnection):
    cls_statement = NamedTupleExaStatement


C = NamedTupleExaConnection(dsn=config.dsn, user=config.user, password=config.password, schema=config.schema)
stmt = C.execute("SELECT * FROM users ORDER BY user_id LIMIT 5")
print(stmt.fetchone())
print(stmt.fetchone())
```

Hope it helps!

wildraid:
By the way, please note that you did not need option `replace=True`. There is an extra check in PyEXASOL to ensure column name uniqueness in result sets. It helps to eliminate potential ambiguous behaviour and helps to maintain integrity of `fetch_dict=True` option.

jabadia:
> However, I've extended [25_overload](https://github.com/badoo/pyexasol/commit/99183cb2588bfb89872684f2d3e05f8cba82e6d3) example to show how you can still use named tuples in your code if you wish to do so.

This is really cool. It absolutely works for me.

> By the way, please note that you did not need option replace=True

I needed it for my specific case, because I had some column names starting with underscore, that made namedtuples fail in some cases (as you already identified as a limitation)

I think that doing this in 'userland' is a much better option, because I recognize that it was my own need the one I was trying to solve.

Thanks!

PR #35, opened 2019-09-24
Add context manager for ExaConnection and ExaStatement

icking:
I'm using pyexasol in some production environment. There is a need to close connections and statements properly to avoid idle ones occupying too much memory on the Exasol cluster.
I added the context manager so that instead of
```python
try:
    conn = pyexasol.connect(...)
    # do something
finally:
    if conn:
        conn.close()
```
we can write
```python
with pyexasol.connect(...) as conn:
    # do something
```

### DIFF
--- a/pyexasol/connection.py
+++ b/pyexasol/connection.py
@@ -155,6 +155,12 @@ class ExaConnection(object):
         self._login()
         self.get_attr()
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, *args):
+        self.close()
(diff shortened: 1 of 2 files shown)


wildraid:
That's a good idea. Please check new version `0.8.0`.

The code is generally the same, but I've used slightly different arguments for `__exit__()` and added  an example with context management.

Hope it helps!

icking:
looks good. I'll close this PR.

PR #41, opened 2019-12-04
Add query_params in DB2Cursor execute method

apeccaud:


wildraid:
Hmm, the the problem is "query_params" in PyEXASOL are quite different from any formats described by DB API 2.0: https://www.python.org/dev/peps/pep-0249/#paramstyle

I am not sure if it should be added to "compatibility" layer, since it's technically not compatible.

Normally it's a good idea to write a small custom connector handling such edge cases rather than relying on DB API 2.0.

apeccaud:
I see. I would like to write a plugin to support exasol in another tool but i need a connection object that is DB API 2.0 compliant. Unfortunately i can't use DB2Connection if i cant execute parametrised queries. What would you recommend?

wildraid:
The main point againt DB API 2.0 is it's incredible inefficiency for "big data" scenarios. Also, it won't allow you to use advanced Exasol features. Everything is just 5-10 times slower than it should be.

If efficiency is not a concern, you may use PyODBC / TurbODBC (https://github.com/blue-yonder/turbodbc) packages instead, which are compatible with DB API 2.0.

apeccaud:
Alright, makes sense. Thank you for your time. Closing the PR

PR #58, opened 2020-08-11
Changed wrong param assert

derolli1976:
Changed assert for EXPORT command


wildraid:
Fixed in https://github.com/badoo/pyexasol/commit/c062c7b22780109453a29d7e5150396f07a66642 instead.

Please note, it also requires some extra changes in documentation, changelog, etc.

Thank you!

PR #65, opened 2021-02-03
Map INTERVAL DAY TO SECOND to datetime.timedelta

jwarlander:
I've recently needed to map INTERVALs from Exasol, and thought it might be a useful contribution to the core PyEXASOL code.

There's some interesting logic in how Python manages negative timedeltas, essentially subtracting an additional day and then always adding back the seconds + microseconds: https://docs.python.org/3/library/datetime.html#timedelta-objects

I didn't see any "regular" tests, so I added some interval checks to the examples code, with an assert that will raise if there's a mismatch.

### DIFF
--- a/examples/00_prepare.py
+++ b/examples/00_prepare.py
@@ -10,6 +10,7 @@ import datetime
 import random
 import string
 import decimal
+import pprint
 
 bool_values = [True, False]
 user_statuses = ['ACTIVE', 'PENDING', 'SUSPENDED', 'DISABLED']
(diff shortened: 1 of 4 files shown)


jwarlander:
Fixed an issue with the `pprint` output comparison; it seems that Python 3.6 and 3.9 generate different output for timedeltas. I changed the test data setup to use `pprint` for generating the expected output, too, so that there is an apples-to-apples comparison.

The only remaining failure above is due to one of the Travis jobs being rate limited by Docker Hub:

```
$ docker run --detach --name exasol --privileged -p 127.0.0.1:8563:$EXASOL_PORT exasol/docker-db:$EXASOL_TAG
Unable to find image 'exasol/docker-db:latest-7.0' locally
latest-7.0: Pulling from exasol/docker-db
anyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.
See 'docker run --help'.
The command "docker run --detach --name exasol --privileged -p 127.0.0.1:8563:$EXASOL_PORT exasol/docker-db:$EXASOL_TAG" failed and exited with 125 during .
```

PR #104, opened 2023-06-22
Migrate project setup to poetry

### LINKED ISSUE #103: Migrate project setup to poetry
## Summary
Migrate to poetry based project setup on `Test-Automation` branch.
```[tasklist]
### Tasks
- [x] Add dev dependencies to pyproject.tom
- [x] Remove old build and project artifacts
- [x] Merge into Test-Automation branch
```

Nicoretti:
fixes #103


littleK0i:
Btw, is there any reason to use poetry specifically?

A simple Python `build` seems to be much shorter and easier to manage.

Recently I've migrated an another project to `build`. This is how it looks:

- https://github.com/littleK0i/SnowDDL/blob/master/setup.cfg
- https://github.com/littleK0i/SnowDDL/blob/master/pyproject.toml

Easy & clean. No issues so far.

https://pypi.org/project/build/

Nicoretti:
Hi @littleK0i,

Thanks for asking; I'll try to give you some background.

> Btw, is there any reason to use poetry specifically?

## Context

The `Exasol Integration-Team` maintains a significant number of Python `projects/repositories`. In order to keep the automation and context switching overhead smaller, we started to harmonize the projects and their "basic" infrastructure regarding setup, build, testing, etc.

If you have a look at our [toolbox](https://exasol.github.io/python-toolbox/index.html) and its [design document](https://exasol.github.io/python-toolbox/design.html), you may get a rough idea. Still, this effort is a work in progress and is being done alongside the regular project work, so there is quite some way to go.

If you dig into the git history of [styleguide](https://github.com/exasol/python-styleguide), you should be able to find the "initial" list of tools and libraries we intended to use. Maybe even some rationale (but not sure if that ever made it into the repository). Developing the tooling and making changes along the way in the projects has shown to be a more efficient (lightweight) process to go for.

## Why did we choose poetry?

* Good dependency resolver
* Integration of dependency locking
* Integration of virtual environments as part of the workflow
* The devs already knew poetry (pros and cons)
* Poetry was already used in a significant number of Python projects
* Changing the build (project management) for "all Python" projects is very costly

## Personal Note

[build](https://pypi.org/project/build/) looks quite interesting. I personally will definitely give it a try. In turn you may also find projects interesting:

* [Flit](https://flit.pypa.io/en/stable/)
* [Hatch](https://hatch.pypa.io/latest/)

Unfortunately, I personally didn't find a PEP-621/518-compliant build & project management system that checks all my boxes yet.

best
Nico

littleK0i:
Thank you for detailed explanation. You listed some good reasons, so it's not a problem.

My final concern is `poetry.lock` file. I suspect, it attempts to "lock" dependencies, which may cause issues if end user is not prepared for that.

For example, this file lists `pandas==2.0.2` which requires Python 3.8+, but what if end user has Python 3.7?

Since I did not use poetry before, I am not sure what might happen with existing users using older versions of dependencies.

Nicoretti:
> My final concern is `poetry.lock` file. I suspect, it attempts to "lock" dependencies, which may cause issues if end user is not prepared for that.
>
> For example, this file lists `pandas==2.0.2` which requires Python 3.8+, but what if end user has Python 3.7?
>
> Since I did not use poetry before, I am not sure what might happen with existing users using older versions of dependencies.

@littleK0i sure that is a legit concern, let me try to bring some light into this.


**TL;DR: Library/Package users will be fine, it is rather for `development` and having reproducible builds/workspaces.**

## Details
I don't think I can do a much better job than the poetry documentation explaining the [.lock file](https://python-poetry.org/docs/basic-usage/#installing-without-poetrylock) and it's semantics. This is why I would suggest you to have a quick look over there (the docs are quite concise).

Still I can add this "practical" approach: If you run `poetry build` and inspect the generated `METADA` (wheel) or `setup.py` (src-dist) you will see that those only obey the constraints from the `pyproject.toml` file.

### Note: 

The Poetry documentation regarding [lock files "As a library developer"](https://python-poetry.org/docs/basic-usage/#as-a-library-developer) basically describes what we are using it for (most of the times).


### Side note regarding Python 3.7:
I think Python 3.7 support can be dropped, due to the fact that it will be [unmaintained in 4 days](https://endoflife.date/python) anyway.

poetry.lock
kaklakariada, line 1: Mark this file as generated in .gitattributes

pyproject.toml
```diff
@@ -0,0 +1,63 @@
...
+    "export"
+]
+
+
+[tool.poetry.urls]
+"Homepage" = "https://www.exasol.com/"
+"Documentation" = "https://github.com/exasol/pyexasol/"
+"Source" = "https://github.com/exasol/pyexasol"
+"Issues" = "https://github.com/exasol/pyexasol/issues"
+"Changelog" = "https://github.com/exasol/pyexasol/CHANGELOG.md"
```
kaklakariada, line 39:
```suggestion
"Changelog" = "https://github.com/exasol/pyexasol/blob/master/CHANGELOG.md"
```
Link is broken

kaklakariada (changes requested):

poetry.lock
Nicoretti: Good point, keep missing that one. :thumbs_up:

CHANGELOG.md
```diff
@@ -2,7 +2,9 @@
 
 ## [Unreleased]
 
-- Switch packaging and project workflows to poetry (internal)
+- Switch packaging and project workflows to poetry (internal0.25.3)
```
kaklakariada, line 5:
```suggestion
- Switch packaging and project workflows to poetry (internal 0.25.3)
```
❓

kaklakariada (approved):

tkilias (approved):

PR #109, opened 2023-11-21
Fix websocket connection isn't properly closed

### LINKED ISSUE #108: 🐞WebSocket connection isn't properly closed in case of process termination
# Summary
The WebSocket connection isn't properly closed using the [disconnect command](https://github.com/exasol/websocket-api/blob/master/docs/commands/disconnectV1.md) at the Exasol WebSocket protocol level when a process is terminated.

## Reproducing the Issue
1. Run the code snippet bellow:

```python
import pyexasol


class ConnectionWrapper:

    def __getattr__(self, item):
        return getattr(self._connection, item)

    def __init__(self, **kwargs):
        self._connection = pyexasol.connect(**kwargs)

    def __del__(self):
        if self._connection:
            self._connection.close()


if __name__ == '__main__':
    dsn = "localhost:8888"
    user = 'sys'
    password = 'exasol'

    con = ConnectionWrapper(dsn=dsn, user=user, password=password)
    res = con.execute("SELECT 42;")
    print(f"{res.fetchall()}")

```
## Expected Behaviour
```python
[(42,)]
```

## Actual Behaviour
```python
[(42,)]
Exception ignored in: <function ConnectionWrapper.__del__ at 0x7fa7c9799990>
Traceback (most recent call last):
  File ".../test_case.py", line 14, in __del__
  File ".../lib/python3.10/site-packages/pyexasol/connection.py", line 456, in close
  File ".../lib/python3.10/site-packages/pyexasol/connection.py", line 524, in req
  File ".../lib/python3.10/site-packages/websocket/_core.py", line 285, in send
  File ".../lib/python3.10/site-packages/websocket/_core.py", line 313, in send_frame
  File ".../lib/python3.10/site-packages/websocket/_core.py", line 527, in _send
  File ".../lib/python3.10/site-packages/websocket/_socket.py", line 172, in send
  File ".../lib/python3.10/site-packages/websocket/_socket.py", line 149, in _send
  File "/usr/lib/python3.10/ssl.py", line 1206, in send
ssl.SSLError: Underlying socket connection gone (_ssl.c:2326)
```

## Root Cause (optional)
<Unknown>

## Context
* The issue only occurs in case of process termination
* The issue has always existed but has been suppressed, see [here](https://github.com/exasol/pyexasol/blob/9ed5b201f8f672e1dad8349317782e5b4fcebf0d/pyexasol/connection.py#L896).
* The impact of the issue should be minor.
   * The underlying `tcp-connection` is disconnected with a `TCP [RST]`, informing the connection server about the disconnect.
   * It should be handled mostly the same, with the exception that the connection may be resumed.

### Hunches
* The problem likely is caused by libraries used by `pyexasol` rather than `pyexasol` itself
* At some point, lower-level code calls `sock.detach()`. It's possible that the underlying socket is being cleaned up by lower-level facilities that can't consider the fact that the socket is still indirectly used or closed (but not yet freed).

### System

**Desktop:**
 - OS: `Linux (Ubuntu 22.10)`
 - Python Version: `3.10.7`
 - pyexasol `0.25.2`

#### Related Issues
* https://github.com/exasol/sqlalchemy-exasol/issues/390

Nicoretti:
Fixes #108

- Add additional information regarding #108 in the code base
- Add known issues

### DIFF
--- a/pyexasol/connection.py
+++ b/pyexasol/connection.py
@@ -45,38 +45,38 @@ class ExaConnection(object):
     threadsafety = 1
 
     def __init__(self
-                 , dsn=None
-                 , user=None
-                 , password=None
-                 , schema=''
-                 , autocommit=constant.DEFAULT_AUTOCOMMIT
-                 , snapshot_transactions=None
-                 , connection_timeout=constant.DEFAULT_CONNECTION_TIMEOUT
-                 , socket_timeout=constant.DEFAULT_SOCKET_TIMEOUT
-                 , query_timeout=constant.DEFAULT_QUERY_TIMEOUT
-                 , compression=False
-                 , encryption=True
-                 , fetch_dict=False
-                 , fetch_mapper=None
-                 , fetch_size_bytes=constant.DEFAULT_FETCH_SIZE_BYTES
-                 , lower_ident=False
-                 , quote_ident=False
-                 , json_lib='json'
-                 , verbose_error=True
-                 , debug=False
-                 , debug_logdir=None
-                 , udf_output_bind_address=None
-                 , udf_output_connect_address=None
-                 , udf_output_dir=None
-                 , http_proxy=None
-                 , client_name=None
-                 , client_version=None
-                 , client_os_username=None
-                 , protocol_version=constant.PROTOCOL_V3
-                 , websocket_sslopt=None
-                 , access_token=None
-                 , refresh_token=None
-                 ):
+            , dsn=None
+            , user=None
+            , password=None
+            , schema=''
+            , autocommit=constant.DEFAULT_AUTOCOMMIT
+            , snapshot_transactions=None
+            , connection_timeout=constant.DEFAULT_CONNECTION_TIMEOUT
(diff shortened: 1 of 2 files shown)


pyexasol/connection.py
```diff
@@ -890,6 +890,29 @@ def __del__(self):
...
+        # the connection still tries using it (see [#108](https://github.com/exasol/pyexasol/issues/108)).
+        #
+        # Despite this problem, having a `__del__` method that performs cleanup remains valuable,
+        # particularly in the context of long-running processes, to prevent resource leaks.
+        #
+        # Possible alternative solutions to address this would be:
+        # * Within the catch, report a clear error message that the user did not close all the connections which were opened.
+        # * Remove the `__del__` function to "force" the user to always address open/close in a clean manner.
+        #
+        # Still, the feedback we received from users so far indicates that neither alternative is desirable for them.
```
tkilias, line 915: Please, add a note that scenario two causes a small left over in the DB which will be removed after a few hours

tkilias (changes requested):

tkilias (approved):

PR #115, opened 2024-04-22
Add itde support

Nicoretti:


.github/workflows/ci-master.yml
```diff
@@ -1,15 +1,20 @@
...
 
 on:
-  workflow_dispatch:
   push:
-    - master
-    - main
-    - "Test-Automation"
+    branches:
+      - "master"
+      - "main"
```
kaklakariada, line 7: Why do you use both master and main branches?

.github/workflows/examples.yml
```diff
@@ -1,88 +1,79 @@
...
+          --docker-db-image-version 7.1.17 --db-mem-size 4GB
 
-      - name: "Wait for Exasol container (Python ${{ matrix.python-version }} Exasol ${{ matrix.exasol-tag }})"
-        run: python examples/_wait_container.py
+      - name: "Wait for Exasol container (UDF extraction)"
+        run: poetry run python examples/_wait_container.py
+
+      - name: "Run examples"
+        run: |
+          poetry run python examples/a00_prepare.py
```
kaklakariada, line 38: Can you list all files in the directory and run them all, instead of listing them?

.github/workflows/ssl_cert.yml
```diff
@@ -1,57 +1,41 @@
...
-    - master
-  pull_request:
-
-  workflow_dispatch:
+  workflow_call:
 
 jobs:
   ssl_cert:
+    name: Build SSL-Cert
+    runs-on: ubuntu-20.04
```
kaklakariada, line 9: Add a comment why the old version is required

noxfile.py
```diff
@@ -36,12 +72,17 @@ def unit_tests(session: Session) -> None:
 
 @nox.session(name="integration-tests", python=False)
 def integration_tests(session: Session) -> None:
-    """Runs the all integration tests"""
-    _integration_tests(session)
+    """Runs the all integration tests
+
+    Args:
+        db_version: version to start, if none no db will be started.
+    """
```
kaklakariada, line 79: Running itests without db does not make sense. Better fail immediately if `db_version` is missing.

kaklakariada (changes requested):

.github/workflows/ci-master.yml
Nicoretti: because currently it is "master" but I assume we will change it in the future

.github/workflows/examples.yml
Nicoretti: Don't wan't to change the behavior of this workflow to much, also the order of execution is important to some tests.

kaklakariada (approved):

PR #122, opened 2024-04-29
feature/convert aXY examples part 1

Nicoretti:


scripts/generate_users.py
```diff
@@ -0,0 +1,98 @@
...
+
+from faker import Faker
+from faker.providers import BaseProvider
+
+
+class UserDataProvider(BaseProvider):
+
+    def date(self):
+        start_date = datetime.date(2018, 1, 1)
+        return start_date + datetime.timedelta(random.randint(1, 365))
```
kaklakariada, line 16: Would it be better to use random numbers with a fixed seed to get reproducible test data? Or does faker do this automatically?

scripts/generate_users.py
```diff
@@ -0,0 +1,98 @@
...
+
+def _create_parser():
+    parser = argparse.ArgumentParser(
+        description="Generate a CSV file containing users",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
+    parser.add_argument(
+        'filename',
+        type=argparse.FileType('w', encoding='utf8'),
+        help='file the resulting csv should be wirtten to.'
```
kaklakariada, line 69:
```suggestion
        help='file the resulting CSV should be written to.'
```

scripts/generate_users.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+    parser.add_argument(
+        '-n', '--count',
+        type=int, default=10000,
+        help='Amount of users to create.'
```
kaklakariada, line 74:
```suggestion
        help='Number of users to create.'
```

test/data/csv/users.csv
kaklakariada, line 1: Mark this as generated in .gitattributes.
Or better: don't check this in but generate the file automatically during tests.

test/integration/conftest.py
```diff
@@ -29,6 +34,135 @@ def connection(dsn, user, password, schema):
...
+        @property
+        def csv_files(self):
+            return self.data_directory.rglob("*.csv")
+
+        def load(self):
+            self._create_dir()
+            self._upload_files()
+            self._import_data()
+
+        def _execute_docker_command(self, command):
```
kaklakariada, line 81: This executes any command, not just docker
```suggestion
        def _execute_command(self, command):
```

test/integration/conftest.py
```diff
@@ -29,6 +34,135 @@ def connection(dsn, user, password, schema):
...
+            self._logger.info("Executing docker command: %s", command)
+            result = subprocess.run(
```
kaklakariada, line 83: Do you need to check the return code or will this throw an exception?

kaklakariada (changes requested):

scripts/generate_users.py
Nicoretti: Previously, the data was generated each time without any assertions for the test. I have now migrated the test data generation into a CLI tool to ensure reusability, if needed. However, the actual test data that is used for testing purposes will be stored in a CSV, and thus, will not change.

test/integration/conftest.py
Nicoretti: The parameter `check=True` enforces a positive return code and throws an error if it's not achieved. Without `check=True`, manual intervention would be necessary.

test/integration/conftest.py
Nicoretti: Thx, I'll change it manually, due to the references.

kaklakariada (approved):

PR #126, opened 2024-05-14
Convert examples from axy to test cases part 4

Nicoretti:
- **Add formatter tests**
- **Add transaction test marker**
- **Add transaction tests**
- **Add test marker for exception tests**
- **Add tests for exceptions**

### DIFF
--- a/test/integration/exceptions_test.py
+++ b/test/integration/exceptions_test.py
@@ -1,0 +1,100 @@
+import pytest
+import pyexasol
+from pyexasol import (
+    ExaConnectionError,
+    ExaAuthError,
+    ExaQueryError,
+    ExaRequestError,
+    ExaRuntimeError,
+    ExaCommunicationError
+)
+
+
+@pytest.mark.exceptions
+def test_bad_dsn(connection):
+    with pytest.raises(ExaConnectionError):
+        pyexasol.connect(dsn="ThisIsBroken")
+
+
+@pytest.mark.exceptions
+@pytest.mark.parametrize("credentials", [
+    {"username": "wrongy"},
+    {"password": "no-tworking"}
+])
+def test_bad_credentails(credentials, dsn, user, password):
+    username = credentials.get("username", user)
+    password = credentials.get("password", password)
+    with pytest.raises(ExaAuthError):
+        pyexasol.connect(
+            dsn=dsn,
+            user=username,
+            password=password
+        )
+
+
+@pytest.mark.exceptions
+def test_invalid_sql(connection):
+    statement = "SELECT1;"
+    with pytest.raises(ExaQueryError):
+        connection.execute(statement)
+
+
+@pytest.mark.exceptions
+def test_read_from_closed_cursor(dsn, user, password, schema):
+    connection = pyexasol.connect(
+        dsn=dsn,
+        user=user,
+        password=password,
+        schema=schema,
+        fetch_size_bytes=1024
+    )
+    query = "SELECT * FROM users;"
+    cursor = connection.execute(query)
+    cursor.fetchone()
+    cursor.close()
+    with pytest.raises(ExaRequestError):
+        cursor.fetchall()
+
+
+@pytest.mark.exceptions
+def test_fetch_without_result_set(connection):
+    cursor = connection.execute("COMMIT;")
+    with pytest.raises(ExaRuntimeError):
+        cursor.fetchone()
+
+
+@pytest.mark.exceptions
+def test_duplicated_select_of_column(connection):
+    query = "SELECT user_name, user_name FROM USERS;"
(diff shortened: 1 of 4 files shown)


test/integration/exceptions_test.py
```diff
@@ -0,0 +1,100 @@
...
+
+@pytest.mark.exceptions
+def test_duplicated_select_of_column(connection):
+    query = "SELECT user_name, user_name FROM USERS;"
+    with pytest.raises(ExaRuntimeError):
+        connection.execute(query)
+
+
+@pytest.mark.exceptions
+def test_attempt_to_Run_query_on_closed_connection(connection):
```
tkilias, line 74:
```suggestion
def test_attempt_to_run_query_on_closed_connection(connection):
```

test/integration/exceptions_test.py
```diff
@@ -0,0 +1,100 @@
...
+        ),
+        pyexasol.connect(
+            dsn=dsn,
+            user=user,
+            password=password,
+            schema=schema,
+        )
+    ]
+
+    connections[1].execute(f"KILL SESSION {connections[0].session_id()}")
```
tkilias, line 97: connections[1] doesn't get closed by the test

test/integration/formatter_test.py
```diff
@@ -0,0 +1,148 @@
...
+    params = {"value": value}
+    query = "SELECT {value};"
+    actual = connection.execute(query, params).fetchval()
+
+    assert expected == actual
+
+
+@pytest.mark.format
+def test_safe_quote_conversion(connection, schema):
+    params = {"table": (schema, "USERS")}
```
tkilias, line 21: here a table with a lower case table name would be good, to check if the quoting actually works

test/integration/formatter_test.py
```diff
@@ -0,0 +1,148 @@
...
+        'Zachary Thompson'
+    ]
+    actual = connection.execute(query, params).fetchcol()
+
+    assert expected == actual
+
+
+@pytest.mark.format
+def test_safe_identifier_conversion(connection):
+    params = {"table": "USERS"}
```
tkilias, line 65: Probably similar as for the quoting, a lowercase table name would be good

test/integration/formatter_test.py
```diff
@@ -0,0 +1,148 @@
...
+
+@pytest.mark.format
+@pytest.mark.parametrize("value", [
+    True, 'True', "TRUE", 'true'
+])
+def test_string_conversion(connection, value):
+    params = {"is_female": value}
+    query = cleandoc("""
+        SELECT user_name FROM USERS
+        WHERE is_female IS {is_female!r}
```
tkilias, line 47: I would almost say, we should actually return the formatted value

test/integration/formatter_test.py
```diff
@@ -0,0 +1,148 @@
...
+def test_safe_decimal_conversion(connection, value, expected):
+    params = {"value": value}
+    query = "SELECT {value!d};"
+    result = connection.execute(query, params)
+    actual = result.fetchval()
+
+    assert expected == actual
+
+
+@ pytest.mark.format
```
tkilias, line 112:
```suggestion
@pytest.mark.format
```

test/integration/transaction_test.py
```diff
@@ -0,0 +1,53 @@
+import pytest
+import pyexasol
+
+
+# For the fetch_dict tests we need to configure the connection accordingly (fetch_dict=True)
```
tkilias, line 5: would it makes sense to add this as docstring

test/integration/transaction_test.py
```diff
@@ -0,0 +1,53 @@
...
+@pytest.fixture
+def connection(dsn, user, password, schema):
+    con = pyexasol.connect(
+        dsn=dsn,
+        user=user,
+        password=password,
+        schema=schema,
+        autocommit=False
```
tkilias, line 13: I don't see fetch_dict=True

tkilias (changes requested):

test/integration/transaction_test.py
Nicoretti: Not needed in this test.

test/integration/transaction_test.py
Nicoretti: No but I see the comment is wrong ... thanks :+1:

test/integration/formatter_test.py
Nicoretti: works with lower case -> changed.

test/integration/formatter_test.py
Nicoretti: This does not work with lower case table name.

tkilias (approved):

PR #128, opened 2024-05-21
Convert examples to tests Part 5

Nicoretti:


test/integration/extensions_test.py
```diff
@@ -0,0 +1,184 @@
...
+            {'name': 'USER_RATING', 'type': 'DECIMAL', 'sql_type': 'DECIMAL(10,5)', 'size': 10, 'scale': 5,
+             'nulls': True,
+             'distribution_key': False, 'default': None, 'comment': None},
+            {'name': 'USER_SCORE', 'type': 'DOUBLE PRECISION', 'sql_type': 'DOUBLE', 'size': 64, 'scale': None,
+             'nulls': True, 'distribution_key': False, 'default': None, 'comment': None},
+            {'name': 'STATUS', 'type': 'VARCHAR', 'sql_type': 'VARCHAR(50) UTF8', 'size': 50, 'scale': None,
+             'nulls': True,
+             'distribution_key': False, 'default': None, 'comment': None}
+        ]
+        actual = connection.ext.get_sys_columns('users')
```
ahsimb, line 69: Is the order certain? Maybe convert to a set?

test/integration/extensions_test.py
```diff
@@ -0,0 +1,184 @@
...
+        assert expected == actual
+
+    def test_get_sys_tables(self, connection):
+        expected = [
+            {'table_name': 'payments', 'table_schema': 'pyexasol_test', 'table_is_virtual': False,
+             'table_has_distribution_key': False, 'table_comment': None},
+            {'table_name': 'users', 'table_schema': 'pyexasol_test', 'table_is_virtual': False,
+             'table_has_distribution_key': False, 'table_comment': None}
+        ]
+        actual = connection.ext.get_sys_tables()
```
ahsimb, line 79: Again, the order.

test/integration/extensions_test.py
```diff
@@ -0,0 +1,184 @@
...
+            'user_id': 'int64',
+            'user_name': 'category',
+            'register_dt': 'datetime64[ns]',
+            'last_visit_ts': 'datetime64[ns]',
+            'is_female': 'category',
+            'user_rating': 'float64',
+            'user_score': 'float64',
+            'status': 'category'
+        }
+        actual = {column: result[column].dtype.name for column in expected}
```
ahsimb, line 122: I assume in this test we are OK with the `result` having more columns than those in the `expected`.

ahsimb (approved):

test/integration/extensions_test.py
Nicoretti: Good point, @ahsimb. I was thinking something similar. Unfortunately, the functions under test return a dictionary based structure, and dictionaries aren't `hashable`. Therefore, it would only be possible if I adjusted the outputted data/format (Which unnecessarily would complicate the test imho).

test/integration/extensions_test.py
Nicoretti: Thanks for mentioning, ordering is here stable to.

test/integration/extensions_test.py
Nicoretti: Yes it was intentional, thanks for mentioning it though, appreciate it.

PR #133, opened 2024-06-12
Add tests for configuration

Nicoretti:
- **Add pytest marker for configuration based tests**
- **Add test for dsn parsing**
- **Add tests for session parameters**
- **Add test for configuration file support**


test/integration/dsn_test.py
```diff
@@ -0,0 +1,120 @@
...
+        ("127.0.0.19", "127.0.0.19", 8564, None),
+        ("127.0.0.18", "127.0.0.18", 8564, None),
+        ("127.0.0.10", "127.0.0.10", 8564, None),
+        ("127.0.0.15", "127.0.0.15", 8564, None),
+        ("127.0.0.17", "127.0.0.17", 8564, None),
+        ("127.0.0.11", "127.0.0.11", 8564, None),
+        ("127.0.0.23", "127.0.0.23", 8563, None),
+        ("127.0.0.12", "127.0.0.12", 8564, None),
+        ("127.0.0.14", "127.0.0.14", 8564, None),
+        ("127.0.0.21", "127.0.0.21", 8563, None),
```
tomuben, line 43: Why is the expected data unordered? Makes reading much more difficult.....

test/integration/dsn_test.py
tomuben, line 1: Does it make sense to add a test for invalid ip addresses/ports, like 255.255.255.255 or 127.0.0.1:70000
?

test/integration/dsn_test.py
Nicoretti: Actually this is the order which was initially returned, anyway I would not want to make assumptions what the used order will be as long as it is parsed correctly.

test/integration/dsn_test.py
Nicoretti: I see your point, generally a good idea I would say but right now I am just trying to convert the existing examples ("tests") to real test to establish a baseline.

tomuben (approved):

PR #138, opened 2024-06-20
Add test for concurrent access on connection

Nicoretti:


test/integration/concurrency_test.py
```diff
@@ -0,0 +1,49 @@
...
+    def __init__(self, connection, timeout):
+        self.connection = connection
+        self.seconds = timeout
+        self.exception = None
+        super().__init__()
+
+    def run(self):
+        try:
+            # run heavy query
+            self.connection.execute("SELECT * FROM users a, users b, users c, payments d")
```
ckunki, line 19: Could the query be simplified?
Or is there a specific reason for querying table `users` 3 times?

ckunki (changes requested):

test/integration/concurrency_test.py
Nicoretti: took this from the original example, see comment above. 
I assume this shall make sure query does take long enough for the second attempt.
Maybe it can be done with a sleep, I'll try.

littleK0i: Previously "sleep" UDFs did not work well with abort_query() call. Usually it caused an entire connection to be terminated.

Nicoretti: @littleK0i from what I have seen >= 7.1, sleep works fine.
With this information which query would you prefer and why?

ckunki (approved):

PR #142, opened 2024-06-25
Add tests for import/export with pandas DataFrames

Nicoretti:
- **Add pandas test marker**
- **Add basic test for pandas import and export**

### DIFF
--- a/test/integration/pandas_test.py
+++ b/test/integration/pandas_test.py
@@ -1,0 +1,114 @@
+import pytest
+import pyexasol
+from inspect import cleandoc
+import pandas as pd
+from pandas import testing as pd_testing
+from operator import itemgetter
+
+
+@pytest.fixture
+def connection_with_compression(dsn, user, password, schema):
+    with pyexasol.connect(
+        dsn=dsn,
+        user=user,
+        password=password,
+        schema=schema,
+        compression=True,
+    ) as connection:
+        yield connection
+
+
+@pytest.fixture
+def empty_table(connection):
+    name = "USER_NAMES"
+    ddl = cleandoc(
+        f"""
+        CREATE OR REPLACE TABLE {name}
+        (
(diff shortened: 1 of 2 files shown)


test/integration/pandas_test.py
```diff
@@ -0,0 +1,114 @@
...
+    "connection", ["connection", "connection_with_compression"], indirect=True
+)
+@pytest.mark.pandas
+def test_export_table_to_pandas(connection, table):
+    table_name, values = table
+
+    expected = pd.DataFrame.from_records(values)
+    actual = connection.export_to_pandas(table_name)
+
+    assert actual.equals(expected)
```
ahsimb, line 71: Maybe it's better to use the `pandas.testing.assert_frame_equal`. Then you will have more options. For example you can disregard the row order, which might be useful.

test/integration/pandas_test.py
```diff
@@ -0,0 +1,114 @@
...
+    expected = pd.DataFrame.from_records(data=[
+        ("Jessica Mccoy", 0),
+        ("Beth James", 1),
+        ("Mrs. Teresa Ryan", 2),
+        ("Tommy Henderson", 3),
+        ("Jessica Christian", 4),
+    ], columns=["USER_NAME", "USER_ID"])
+    actual = connection.export_to_pandas(query)
+
+    assert actual.equals(expected)
```
ahsimb, line 91: Same as before.

ahsimb (approved):

test/integration/pandas_test.py
Nicoretti: Hi @ahsimb,

Good point, I must admit I had it in before but for me this version seemed to be simpler to use :sweat_smile: likely because I am not so familiar with pandas and it wasn't immediately obvious which settings to use, because the default did not seem to be sufficient in my case.

PR #143, opened 2024-06-26
Add tests for ETL functionality

Nicoretti:


test/integration/etl_test.py
```diff
@@ -0,0 +1,90 @@
...
+
+    ddl = f"DROP TABLE IF EXISTS {name};"
+    connection.execute(ddl)
+    connection.commit()
+
+
+@pytest.mark.etl
+@pytest.mark.parametrize(
+    "params,expected",
+    [
```
tkilias, line 53: The parameters are partially hard to read, because they are so diverse

tkilias: And do you need expected? It seems always the same

test/integration/etl_test.py
```diff
@@ -0,0 +1,90 @@
...
+
+    query = cleandoc(
+        f"""
+        SELECT * FROM {table}
+        WHERE USER_ID NOT IN (SELECT USER_ID from {import_table});
+        """
+    )
+    result = connection.execute(query)
+
+    expected = []
```
tkilias, line 87: Is it really a good idea to actually do not import and export any data

test/integration/export_test.py
```diff
@@ -0,0 +1,121 @@
...
+def swaped_data(data):
+    yield [(lastname, firstname) for firstname, lastname in data]
+
+
+@pytest.fixture
+def table(connection, empty_table, data):
+    name = empty_table
+    stmt = "INSERT INTO {table} VALUES ({{col1}}, {{col2}});"
+    for col1, col2 in data:
+        connection.execute(stmt.format(table=name), {"col1": col1, "col2": col2})
```
tkilias, line 55: The format would be usually not the best idea, but can be ignored here.

test/integration/export_test.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+@pytest.fixture
+def expected_csv(csv_dialect):
+    def create_csv(table, data, **params):
+        csvfile = io.StringIO()
+        if "with_column_names" in params and params["with_column_names"]:
+            data = [("FIRSTNAME", "LASTNAME")] + data
+        writer = csv.writer(csvfile, csv_dialect)
+        writer.writerows(data)
```
tkilias, line 83: Does the csv.writer needs to be closed?

test/integration/export_test.py
```diff
@@ -0,0 +1,121 @@
...
+    connection.commit()
+
+
+@pytest.fixture
+def data(faker):
+    yield [(faker.first_name(), faker.last_name()) for _ in range(0, 10)]
+
+
+@pytest.fixture
+def swaped_data(data):
```
tkilias, line 46: I am not seeing this fixture used. Am I blind?

test/integration/import_test.py
```diff
@@ -0,0 +1,106 @@
...
+    file = tmp_path / "names.csv"
+    with open(file, "w", newline="") as csvfile:
+        writer = csv.writer(csvfile, dialect="unix")
+        for row in data:
+            writer.writerow(row)
+    yield file
+
+
+@pytest.mark.etl
+def test_import_with_custom_csv_format(connection, empty_table, csv_file, data):
```
tkilias, line 60: How is this a custom csv format, it seems to unix dialect , is this something special?

tkilias (changes requested):

test/integration/export_test.py
Nicoretti: AFIK no, it takes a file like object as it's first parameter. From my understanding the one providing the file like object is responsible to close the provided file-like object afterwards (later).

test/integration/import_test.py
Nicoretti: No maybe the name is not ideal, I don't recall why I'll rename it.

test/integration/etl_test.py
Nicoretti: The expected parameter should be removed. In fact, a full export and import is conducted. 

Here is the concept of the test:

1. Export: `USERS` Table to `CSV File`
2. Import: `CSV File` to `USERS_IMPORT` Table
3. Verify that the sets/tables are identical

tkilias: a table is the table with data and input_table the empty target table

tkilias (approved):

PR #145, opened 2024-07-02
Add dbapi2 compatibility driver

Nicoretti:
Migrate dbapi2 compliant shim/driver from [SQLA](https://github.com/exasol/sqlalchemy-exasol/tree/master/exasol/driver).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Add dbapi2 compliant driver interface `exasol.driver.websocket` ontop of pyexasol
 - Switch packaging and project workflows to poetry
 - Drop support for python 3.7
 - Drop support for Exasol 6.x
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- Add dbapi2 compliant driver interface `exasol.driver.websocket` ontop of pyexasol
+- Added dbapi2 compliant driver interface `exasol.driver.websocket` ontop of pyexasol
 - Switch packaging and project workflows to poetry
 - Drop support for python 3.7
 - Drop support for Exasol 6.x
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,11 @@
 ## [Unreleased]
 
 - Added dbapi2 compliant driver interface `exasol.driver.websocket` ontop of pyexasol
-- Switch packaging and project workflows to poetry
-- Drop support for python 3.7
-- Drop support for Exasol 6.x
-- Drop support for Exasol 7.0.x
-- Relock dependencies (Internal)
+- Switched packaging and project workflow to poetry
+- Droped support for python 3.7
+- Droped support for Exasol 6.x
+- Droped support for Exasol 7.0.x
+- Relocked dependencies (Internal)
 
 ## [0.25.2] - 2023-03-14
 
--- a/docs/DBAPI_COMPAT.md
+++ b/docs/DBAPI_COMPAT.md
@@ -21,6 +21,9 @@ with connection.cursor() as cursor:
 #### TurboODBC
 [TurboODBC](https://github.com/blue-yonder/turbodbc) offers an alternative ODBC-based, DBAPI2-compatible driver, which supports the Exasol database.
 
+#### Pyodbc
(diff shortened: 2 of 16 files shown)


littleK0i:
Not sure if it is a good idea. Historically pyexasol was slightly incompatible with DB API 2.0 on purpose. It forced users to look at documentation for a few minutes instead of mindlessly copying incorrect advice from StackOverflow.

Main issues with DB API 2.0:

- designed a long time ago;
- designed for OLTP databases running on single node;

Fundamentally implementing it would add more harm than good.

Nicoretti:
Hi @littleK0i,

I understand and share your concern. That is why I made the shim a separate package. It won't even be part of the `pyexasol` namespace. If you think the docs may lure some folks into just copying the example without reading it, we could put the part on the alternatives at the end of the document.


## Why we are doing this?
We have seen use cases and received requests in the past where DBAPI2 compliance was desired and reasonable, or the only way to integrate with other frameworks. Therefore, we believe we should allow users to achieve this. Although there is `turbodbc`, it is not a viable alternative in all cases, particularly since it is ODBC-based.

In the end, it is up to the user and their needs. I do not see why we should patronize the library users. We should make the right things easy and the potentially harmful ones harder, but ultimately the user should be able to decide because they should know what they need.


> Fundamentally implementing it would do more harm than good. 

Having worked mainly with C++ in the past (8+ years), I totally understand where you're coming from.
However, we strongly believe that this change will be beneficial for some users.

Nevertheless, any suggestions and discussions are welcome so that we can enable the user to make a well-informed decision about the potential tradeoffs at hand.

I hope this helps to better understand the background of this change. 
Looking forward to getting further feedback from your side.

best
Nico

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Add dbapi2 compliant driver interface `exasol.driver.websocket` ontop of pyexasol
```
ckunki, line 5:
```suggestion
- Added dbapi2 compliant driver interface `exasol.driver.websocket` ontop of pyexasol
```

pyexasol/warnings.py
```diff
@@ -0,0 +1,6 @@
+class PyexasolWarning(UserWarning):
+    """Base class for all warnings emited by pyexasol."""
```
ckunki, line 2:
```suggestion
    """Base class for all warnings emitted by pyexasol."""
```

exasol/driver/websocket/_errors.py
```diff
@@ -0,0 +1,76 @@
+"""
+This module provides contains `PEP-249`_ compliant DBAPI exceptions.
```
ckunki, line 2:
```suggestion
This module provides `PEP-249`_ compliant DBAPI exceptions.
```

exasol/driver/websocket/_errors.py
```diff
@@ -0,0 +1,76 @@
...
+
+class Warning(Exception):  # Required by spec. pylint: disable=W0622
+    """
+    Exception raised for important warnings like data truncations while inserting, etc.
+    """
+
+
+class Error(Exception):
+    """
+    Exception that is the base class of all other error exceptions.
```
ckunki, line 17:
```suggestion
    Base class of all other error exceptions.
```

ckunki (changes requested):

docs/DBAPI_COMPAT.md
```diff
@@ -1,8 +1,25 @@
...
+```python
+from exasol.driver.websocket.dbapi2 import connect
+
+connection = connect(dsn='', username='sys', password='exasol', schema='TEST')
+
+with connection.cursor() as cursor:
+    cursor.execute("SELECT 1;")
+```
+
+#### TurboODBC
```
tkilias, line 21: is it only turboodbc, or also pyodbc?

Nicoretti: pyodbc should work too I guess

Nicoretti: Should we add it?

tkilias: > pyodbc is an open source Python module that makes accessing ODBC databases simple. It implements the DB API 2.0 specification but is packed with even more ...
https://pypi.org/project/pyodbc/

it does, and yes please add it

tkilias (approved):

ckunki (approved):

Nicoretti:
@littleK0i

I'll merge this PR for now. If you have any further feedback or feel that I may have been too quick with this change, please feel free to contact me directly via email. Your feedback is always highly appreciated and welcome.

Best,
Nico

PR #149, opened 2024-07-09
Add tests for http transport errors

Nicoretti:


test/integration/http_transport_errors_test.py
```diff
@@ -0,0 +1,83 @@
...
+    yield "SELECT * FROM DoesNotExist LIMIT 1000"
+
+
+@pytest.mark.etl
+@pytest.mark.exceptions
+def test_sql_error(connection, dev_null, invalid_statement):
+    def export_cb(pipe, dst, **kwargs):
+        shutil.copyfileobj(pipe, dev_null)
+
+    with pytest.raises(pyexasol.exceptions.ExaQueryError):
```
kaklakariada, line 30: Maybe also assert exception message?

test/integration/http_transport_errors_test.py
```diff
@@ -0,0 +1,83 @@
...
+    def export_cb(pipe, dst, **kwargs):
+        raise Exception(error_msg)
+
+    with pytest.raises(Exception) as exec_info:
+        connection.export_to_callback(export_cb, None, statement)
+
+    expected = error_msg
+    actual = f"{exec_info.value}"
+
+    assert actual == expected
```
kaklakariada, line 61:
```suggestion
    with pytest.raises(Exception, matches=error_msg):
        connection.export_to_callback(export_cb, None, statement)
```

Not sure if this works ;)

kaklakariada (approved):

test/integration/http_transport_errors_test.py
Nicoretti: In my opinion, there's occasional merit in verifying the actual Exception message, alongside the current test. Be aware, though, this could lead to a more fragile test. In this instance, I didn't see the need to confirm the actual exception message. Thanks for pointing it out though :+1: .

kaklakariada (approved):

PR #160, opened 2024-12-06
Add support for standard exasol-toolbox nox tasks

Nicoretti:


noxconfig.py
```diff
@@ -6,6 +6,44 @@
...
+from exasol.toolbox.nox.plugin import hookimpl
+
+
+class StartDB:
+
+    @hookimpl
+    def pre_integration_tests_hook(self, session, config, context):
+        port = context.get("port", 8563)
+        db_version = context.get("db_version", "7.1.17")
+        session.run(
```
tkilias, line 18: Is the itde API maybe better it also provides a cleanup function

Nicoretti: For now, I would leave it as is because I wanted a straightforward transition from the existing state to the new one. In this case, it was used exactly the same, just called differently. In the future, it might make sense to use the API, though.

tkilias: Ok, perfectly fine

tkilias (approved):

Issue #117, opened 2024-04-23
✨ Create a `1.0.0` release


Nicoretti:
# Summary
Create a `1.0.0` release for the project oncthat the test and release infrastructure is in place.

# Details
As the major version number is currently less than 1, any non-major version change could potentially break the API. This makes it difficult to communicate any breaking changes to users or build systems that rely on the version number. However, the API is widely used and stable, so it's time to communicate its maturity to our users by releasing version 1.0.0.

## References
- See https://semver.org/lang/en/ for more information about versioning and communicating changes to users.

PR #177, opened 2025-03-04
Security/512 enforce strict certificate validation

ArBridgeman:


ArBridgeman:
These todos are written based on the assumption that initial implementation is not changed. They would need to be reworded pending changes:

- [ ] Check github CI runs -> looks like examples with 3.11 are failing
- [ ] Discuss which documentation & examples should be updated & how (i.e. do we want with security-minded effort everywhere or without)
   - [ ] README
   - [ ] doc
   - [ ] examples
   - [ ] pyexasol_utilities
   - [ ] performance
- [ ] Test for validating encryption + certificate as default works as expected
- [ ] Anything else as major release change to mind?

pyexasol/connection.py
```diff
@@ -1206,16 +1206,16 @@ def _get_ws_options(self):
         }
 
         if self.options["encryption"]:
```
ArBridgeman, line 1208: Move info to a more apparent level (or change implementation)? As in `_get_ws_options` of class, it's not shown in the generated documentation. Could set in `__init__` or more directly expose.

pyexasol/connection.py
```diff
@@ -1206,16 +1206,16 @@ def _get_ws_options(self):
...
-            # Use "websocket_sslopt" argument to set custom SSL options, which are passed directly to WebSocket client constructor
-            if self.options["websocket_sslopt"]:
-                options["sslopt"] = self.options["websocket_sslopt"]
-            # OpenID tokens are normally used for Exasol SAAS
-            # Strict certificate verification is enabled by default
-            elif self.options["access_token"] or self.options["refresh_token"]:
```
ArBridgeman, line 1214: needs to stay

ArBridgeman: @tkilias please check

doc/changes/unreleased.md
```diff
@@ -7,7 +7,8 @@
 
 ## ⚒️ Refactorings
 
-* Reformated entire code base with `black` and `isort`
+* Reformatted entire code base with `black` and `isort`
+* Modified `ExaConnection` so that default is encryption with strict certification validation
```
ArBridgeman, line 11: not refactoring -> feature
add comment at top with breaking change and move to version 1.
-> see guidelines for

ArBridgeman:
@tkilias is there anything related to https://github.com/exasol/pyexasol/issues/117 that we need to be aware of/do before a `1.0.0` release?

examples/_wait_container.py
ArBridgeman, line 1: In CI and/or wait script, we should add a break condition -> while True

ArBridgeman: moved to https://github.com/exasol/pyexasol/issues/178

tkilias:
> @tkilias is there anything related to https://github.com/exasol/pyexasol/issues/117 that we need to be aware of/do before a `1.0.0` release?

I think,  the ticket asks for checking the versions of the dependencies. At the moment,  we allow all versions that is maybe to broad, but on the other hand we can't to narrow either.  We need to find proper constraints for the versions.

doc/changes/unreleased.md
```diff
@@ -1,13 +1,35 @@
 # Unreleased
 
+## Summary
+
+From PyExasol version ``1.0.0``, the default behavior has been changed to use strict 
+certificate verification for encrypted connections (default, ``encryption=True``) in
```
tkilias, line 6: Encryption is probably already enabled by default, because the DB only accepts those connections. However, the default websock_sslopts have changed.

ArBridgeman:
Closing to create separate issue/branch for public visibility of fix.
https://github.com/exasol/pyexasol/pull/180 and https://github.com/exasol/pyexasol/issues/179

PR #184, opened 2025-03-20
Feature/181 run tests also with strict certificate

### LINKED ISSUE #181: ✨ Modify PyExasol tests to run with strict certificate
# Summary
With the resolution of https://github.com/exasol/pyexasol/issues/179, we've changed the default to be with strict certificate verification. However, in our CI pipelines, we are running integration tests only against ssl.CERT_NONE. We should adapt our integration tests to run against both scenarios. 

## References
* https://docs.pytest.org/en/stable/example/parametrize.html#parametrization-of-optional-implementations-imports
* https://github.com/exasol/pyexasol/blob/master/.github/workflows/ssl_cert.yml

## Task
This can be accomplished with some modifications: (1) copy crt  from docker container & (2) adapt `websocket_sslopt ` pytest fixture so that every test is run twice: (1) ssl.CERT_NONE and (2) ssl.CERT_REQUIRED with crt. As (2) is now our default, we should run it as well and (1) would be expected for Docker & some "on-premises" setups.

ArBridgeman:
Closes #181

### DIFF
--- a/.github/workflows/slow-checks.yml
+++ b/.github/workflows/slow-checks.yml
@@ -38,6 +38,9 @@ jobs:
       - name: "Install all extras"
         run: poetry install --all-extras
 
+      - name: "Setup custom host"
+        run: echo "127.0.0.1 exasol-test-database" | sudo tee -a /etc/hosts
+
       - name: Run Tests and Collect Coverage
         run: poetry run nox -s test:integration -- -- --coverage --db-version ${{ matrix.exasol-version }}
 
--- a/doc/developer_guide.rst
+++ b/doc/developer_guide.rst
@@ -47,6 +47,13 @@ Integration Tests
 
    As a Docker container with a test databases needs to be started for the integration tests, it may take a bit before the tests themselves start executing. After the tests have been run, the database will be shut down again.
 
+   If you are running Docker inside of a VM and are running the integration tests with `ssl.CERT_REQUIRED`, inside your VM map the `exasol-test-database` to the associated IP address.
+   This mapping is required due to how the certificate was created & for hostname resolution.
+
+    .. code-block:: shell
+
+        echo "127.0.0.1 exasol-test-database" | sudo tee -a /etc/hosts
+
 .. code-block:: shell
 
     nox -s test:integration
--- a/noxconfig.py
+++ b/noxconfig.py
@@ -5,39 +5,60 @@ from pathlib import Path
 from typing import Iterable
 
 from exasol.toolbox.nox.plugin import hookimpl
+from nox import Session
 
+DEFAULT_PORT = 8563
 DEFAULT_DB_VERSION = "8.31.0"
+CONTAINER_SUFFIX = "test"
+CONTAINER_NAME = f"db_container_{CONTAINER_SUFFIX}"
+
+
+def start_test_db(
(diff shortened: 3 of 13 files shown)


.github/workflows/slow-checks.yml
```diff
@@ -38,6 +38,9 @@ jobs:
       - name: "Install all extras"
         run: poetry install --all-extras
 
+      - name: "Setup custom host"
```
tkilias, line 41: Uh, yeah you probably need this now

tkilias: That gets fancy when we want to press into pytest fixtures, but for the moment not an issue

noxconfig.py
```diff
@@ -5,39 +5,60 @@
...
+        "--db-mem-size",
+        "4GB",
+        external=True,
+    )
+    session.run(
+        "docker",
+        "cp",
+        f"{CONTAINER_NAME}:/certificates/rootCA.crt",
+        f"{Path(__file__).parent}/test/data/rootCA.crt",
+    )
```
tomuben, line 41: Maybe, move the copy cmd to a pytest fixture, which copies the certificate from the docker container into a temporary folder, which will be removed after the tests finished. just an idea....

doc/developer_guide.rst
```diff
@@ -47,6 +47,13 @@ Integration Tests
 
    As a Docker container with a test databases needs to be started for the integration tests, it may take a bit before the tests themselves start executing. After the tests have been run, the database will be shut down again.
 
+   If you are running Docker inside of a VM and are running the integration tests with `ssl.CERT_REQUIRED`, inside your VM map the `exasol-test-database` to the associated IP address.
+   This mapping is required due to how the certificate was created & for hostname resolution.
+
+    .. code-block:: shell
+
+        echo "127.0.0.1 exasol-test-database" | sudo tee -a /etc/hosts
+
```
tomuben, line 56: Another solution would be to let the integration tests spawn a new docker container which prepares this environment.....

tkilias: but then we would need to run the tests in this docker container

tkilias: I think, for the moment, we stay with the current variant, but we should have a ticket for a better solution, maybe the solution is also in itde, that we make it configurable what name is in the certificate @ArBridgeman please create a ticket that we think about how to improve this

ArBridgeman: Feel free to edit/comment https://github.com/exasol/integration-test-docker-environment/issues/462

test/integration/conftest.py
```diff
@@ -11,12 +11,28 @@
 import pytest
 
 import pyexasol
+from noxconfig import CONTAINER_NAME
```
tomuben, line 14: Ah, hm, I didn't think about this dependency when I made the suggestion.
@tkilias , do you think this is problematic?

ArBridgeman: I could alternately add a fixture for it with the fixed text. atm, it is also hard-coded separately for the `DockerDataLoader` later in the `conftest.py`.

tkilias: hm, importing from the noxconfig, it probably can cause issues and make the test potentially depend on your current working directory

tkilias: @ArBridgeman yes, the fixture is safer

tomuben (approved):

tkilias (approved):

PR #203, opened 2025-06-24
Feature/194 handle certificate for import and export

### LINKED ISSUE #194: ✨ Handle certificate verification affecting import & export query
# Summary
Pyexasol's import & export queries use a self-signed certificate, which is not sufficient for passing as a globally trusted certificate. The queries should be modified to properly handle this with later DB versions.

Two possible ways to resolve this are:
1. Add a parameter in pyexasol, such as `IGNORE_CERTIFICATE`, to bypass verification.
2. Support the addition of pyexasol’s certificate as a PUBLIC KEY, defined in the format PUBLIC KEY 'sha256//abcde256', to allow proper verification handling. -> this is only supported by 8.32+

## Background & Context
We are planning to enable certificate verification as the default behavior for database Import/Export queries. This change will cause queries to fail if certificate verification is unsuccessful.

In the case of pyexasol, the import query is built here:
https://github.com/exasol/pyexasol/blob/f120ea4592cf759f24132352bacb409d84b2fbae/pyexasol/http_transport.py#L204
For the export query:
https://github.com/exasol/pyexasol/blob/f120ea4592cf759f24132352bacb409d84b2fbae/pyexasol/http_transport.py#L139

With this change, import/export queries will fail because Pyexasol currently uses a self-signed certificate, which cannot pass verification as a globally trusted certificate.

ArBridgeman:
closes #194


littleK0i:
Using version checks is a bit outdated and fragile approach.

As a better alternative, server may return list of active "feature flags" defining state of important features. These feature flags must be checked instead of version number.

It allows features to be enabled or disabled not only for specific DB version, but also depending on settings, depending on specific users, etc. It is also resistant to versioning schema changes.

---

Alternatively, Exasol already has `protocolVersion`. Maybe it can be used instead. If database supports protocolVersion=5, it might be safe to assume it supports public key in IMPORT / EXPORT as well.

pyexasol/http_transport.py
```diff
@@ -6,6 +9,245 @@
...
+            self._column_separator,
+            self._column_delimiter,
+        ]
+        return self._get_query_str(query_lines)
+
+    @staticmethod
+    def load_from_dict(
+        connection: ExaConnection, compression: bool, params: dict
+    ) -> ImportQuery:
+        return ImportQuery(connection=connection, compression=compression, **params)
```
ArBridgeman, line 180: note that if folks pass in other values to params than expected that the behavior has changed -> before we just pulled values we expected from the params dict and now we pass them into a class, i.e.. they have "dummy", it will now raise an exception saying "dummy" isn't a valid value

ckunki (commented):
:heart: First of all: thank you so much for the excellent description in the ticket #194 which really warped me a big step towards understanding what the current PR is about.

pyexasol/http_transport.py
```diff
@@ -193,63 +345,19 @@ def __init__(self, connection, compression, table, import_params):
...
+        ).build_query(source=table_ident, exa_address_list=self.exa_address_list)
+        self.connection.execute(import_query)
 
 
 class ExaHttpThread(threading.Thread):
     """
-    HTTP communication and compression / decompression is offloaded to separate thread
-    Thread can be used instead of subprocess when compatibility is an issue
+    HTTP communication and compression / decompression is offloaded to a separate
+    thread. A thread can be used instead of a subprocess when compatibility is an issue.
```
ckunki, line 357: What is the intention of the last sentence?
Justify the design decision to use a thread? 
Is compatibility an issue, here? 
In which regard?

pyexasol/http_transport.py
```diff
@@ -508,45 +627,90 @@ def set_sock_opts(self):
...
+                    encoding=serialization.Encoding.PEM,
+                    format=serialization.PrivateFormat.TraditionalOpenSSL,
+                    encryption_algorithm=serialization.NoEncryption(),
+                )
+            )
             key_file.close()
 
-            context = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
-            context.verify_mode = ssl.CERT_NONE
+            context = SSLContext(PROTOCOL_TLS_SERVER)
```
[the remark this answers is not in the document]
ckunki: I agree.

doc/changes/unreleased.md
```diff
@@ -2,32 +2,59 @@
...
-simply upgrading to version ``1.0.0`` is likely to lead to breaking changes with error
-messages like: ``[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed``.
+simply upgrading to version `1.0.0` is likely to lead to breaking changes with error
+messages like: `[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed`.
 
 Prior to the upgrade:
 1. Please determine which encryption and security measures are appropriate for your
-organization by reading through ``doc/user_guide/encryption.rst``.
-2. If needed, update your usage of ``pyexasol.connect(...)`` & ``ExaConnection`` to
+organization by reading through `doc/user_guide/encryption.rst`.
```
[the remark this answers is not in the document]
ArBridgeman: Good point, we could put the links in, and I don't think it'd be a huge risk, as the link checker is a few minor changes away from rolling out :crossed_fingers:

test/unit/http_transport_test.py
```diff
@@ -0,0 +1,368 @@
...
+        assert sql_query._build_columns_str() == expected
+
+    @staticmethod
+    @pytest.mark.parametrize(
+        "csv_cols,expected",
+        [
+            pytest.param(None, "", id="none_specified_returns_empty_string"),
+            pytest.param(["1..2"], "(1..2)", id="values_specified_returns_value"),
+        ],
+    )
```
ckunki, line 59: Please add test cases for 
* more supported variants, e.g. "123", `FORMAT='abc'`

test/unit/http_transport_test.py
```diff
@@ -0,0 +1,368 @@
...
+        sql_query.encoding = encoding
+        assert sql_query._encoding == expected
+
+    @staticmethod
+    @pytest.mark.parametrize(
+        "compression,file_ext,expected",
+        [
+            pytest.param(True, None, "gz", id="no_format_maps_to_gz"),
+            pytest.param(False, None, "csv", id="no_format_maps_to_csv"),
+            pytest.param(True, "gz", "gz", id="gz_accepted"),
```
ckunki, line 204:
```suggestion
            pytest.param(True, None, "gz", id="compressed_defaults_to_format_gz"),
            pytest.param(False, None, "csv", id="uncompressed_defaults_to_format_csv"),
            pytest.param(True, "gz", "gz", id="format_gz_accepted"),
```

test/unit/http_transport_test.py
```diff
@@ -0,0 +1,368 @@
...
+    def test_row_separator(sql_query, row_separator, expected):
+        sql_query.row_separator = row_separator
+        assert sql_query._row_separator == expected
+
+
+class TestImportQuery:
+    @staticmethod
+    def test_build_query(import_sql_query, connection):
+        result = import_sql_query.build_query(
+            source="TABLE",
```
ckunki, line 253: "import into" would sound more natural with "target" or "table" rather than "source".

test/unit/http_transport_test.py
```diff
@@ -0,0 +1,368 @@
...
+            (
+                ["LASTNAME", "FIRSTNAME"],
+                'IMPORT INTO TABLE("LASTNAME","FIRSTNAME") FROM CSV',
+            ),
+            (None, "IMPORT INTO TABLE FROM CSV"),
+        ],
+    )
+    def test_get_import(import_sql_query, columns, expected):
+        import_sql_query.columns = columns
+        result = import_sql_query._get_import(source="TABLE")
```
ckunki, line 282: Suggestion: replace `source` by `table` or `target`.

test/unit/http_transport_test.py
```diff
@@ -0,0 +1,368 @@
...
+        assert result == expected
+
+    @staticmethod
+    @pytest.mark.parametrize(
+        "skip,expected", [("1", "SKIP = 1"), (1, "SKIP = 1"), (None, None)]
```
ckunki, line 287: maybe add a test case with `"2"` or `2`?

pyexasol/http_transport.py
ArBridgeman: This was already handled in https://github.com/exasol/pyexasol/pull/205, so i'll close it here.

pyexasol/http_transport.py
```diff
@@ -6,6 +9,246 @@
...
+        prefix = self._url_prefix
+
+        csv_cols = self._build_csv_cols()
+        files = []
+        for i, exa_address in enumerate(exa_address_list):
+            statement = f"AT '{prefix}{exa_address}'"
+            if self._requires_tls_public_key():
+                sha256_public_key = self._extract_public_key(exa_address)
+                statement += f" PUBLIC KEY 'sha256//{sha256_public_key}'"
+            statement += f" FILE '{str(i).rjust(3, '0')}.{file_ext}'{csv_cols}"
```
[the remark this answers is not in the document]
ArBridgeman: hmm...i wouldn't think so, but it comes from the previous work. as this PR is rather large, i'd recommend we discuss if this is useful to change, and if so, to create a new issue to address it.

cc: @tkilias

pyexasol/http_transport.py
```diff
@@ -6,6 +9,246 @@
...
+
+    @property
+    def _file_ext(self) -> str:
+        if not self.format:
+            if self.compression:
+                return "gz"
+            return "csv"
+        if self.format not in ("gz", "bz2", "zip"):
+            raise ValueError(f"Unsupported compression format: {self.format}")
+        return self.format
```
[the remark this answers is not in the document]
ArBridgeman: this one I'd rather not change.

pyexasol/http_transport.py
```diff
@@ -508,45 +627,90 @@ def set_sock_opts(self):
...
+        from tempfile import TemporaryDirectory
 
-        from OpenSSL import crypto
+        from cryptography import x509
+        from cryptography.hazmat.primitives import (
+            hashes,
+            serialization,
+        )
+        from cryptography.hazmat.primitives.asymmetric import rsa
+        from cryptography.x509.oid import NameOID
```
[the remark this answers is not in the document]
tkilias: very good

test/unit/http_transport_test.py
```diff
@@ -0,0 +1,375 @@
...
+    def test_build_query(import_sql_query, connection):
+        result = import_sql_query.build_query(
+            table="TABLE",
+            exa_address_list=[
+                "127.18.0.2:8364/YHistZoLhU9+FKoSEHHbNGtC/Ee4KT75DDBO+s5OG8o="
+            ],
+        )
+        assert (
+            result
+            == "IMPORT INTO TABLE FROM CSV\nAT 'https://127.18.0.2:8364/YHistZoLhU9+FKoSEHHbNGtC/Ee4KT75DDBO+s5OG8o=' PUBLIC KEY 'sha256//YHistZoLhU9+FKoSEHHbNGtC/Ee4KT75DDBO+s5OG8o=' FILE '000.gz'"
```
tkilias, line 266: hm, yeah, not sure if the public be should be part of the url

pyexasol/http_transport.py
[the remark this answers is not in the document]
ckunki: OK

ckunki (approved):
LGTM

tkilias (approved):

ArBridgeman:
> Using version checks is a bit outdated and fragile approach.
> 
> As a better alternative, server may return list of active "feature flags" defining state of important features. These feature flags must be checked instead of version number.
> 
> It allows features to be enabled or disabled not only for specific DB version, but also depending on settings, depending on specific users, etc. It is also resistant to versioning schema changes.
> 
> Alternatively, Exasol already has `protocolVersion`. Maybe it can be used instead. If database supports protocolVersion=5, it might be safe to assume it supports public key in IMPORT / EXPORT as well.

Thanks for bringing this to our attention. After coordinating with those involved, it's been clarified that, for this case, the DB version would be the most stable solution.

PR #209, opened 2025-07-08
Feature/208 add import from parquet

### LINKED ISSUE #208: Add parquet import
# Summary
A database is only as useful as the data stored in it. Parquet is a very popular data storage format. We want to enable users to work with their Parquet data in Exasol.

# Details
**Out of Scope**
- Import from any other storage than a local (or locally mounted) filesystem
- Hierarchical data import

# Task(s)

- [ ] pyexasol imports flat table data from a Parquet file from the local file system
- [ ] documentation updated
- [ ] tests added

ArBridgeman:
closes #208 

Performance checks & table creation, when not existing yet, with inferring the data types will come in subsequent issues/PRs.

### DIFF
--- a/pyexasol/connection.py
+++ b/pyexasol/connection.py
@@ -565,6 +565,7 @@ class ExaConnection:
         )
 
     def export_to_polars(
+        self,
         query_or_table: str,
         query_params: Optional[dict] = None,
         callback_params: Optional[dict] = None,


pyexasol/callback.py
```diff
@@ -74,33 +76,56 @@ def import_from_pandas(pipe, src, **kwargs):
...
-            header=False,
-            index=False,
-            line_terminator="\n",
-            quoting=csv.QUOTE_NONNUMERIC,
-            **kwargs
-        )
+    if not isinstance(source, Path):
+        raise ValueError(f"source {source} is not a `pathlib.Path`")
+
+    matching_files = list(source.parent.glob(source.name))
```
tkilias, line 116: that is no how you should use glob, this is an implicit glob, I think, the more clear interface would be, Path, str where we apply glob and iterable of Path

littleK0i:
Well... parquet decompression should be implemented on server side, with parallelism, vectorized reading, full cloud storage support and other optimisations. Nowadays meaningful data in parquet format is unlikely to be stored "locally".

Doing it on the client side and single-threaded manner is fundamentally pointless from the performance perspective. If you do it purely to check the box for "parquet support", it is not enough. This approach won't pass any serious PoC.

pyexasol/callback.py
```diff
@@ -74,33 +78,88 @@ def import_from_pandas(pipe, src, **kwargs):
...
+
+    if not (parquet_files := get_parquet_files(source)):
+        raise ValueError(f"source {source} does not match any files")
+
+    for file in parquet_files:
+        parquet_file = parquet.ParquetFile(file)
+        num_row_groups = parquet_file.num_row_groups
+
+        for i in range(num_row_groups):
+            row_group_table = parquet_file.read_row_group(i)
```
tkilias, line 154: We might need to think about if a user wants to select specific columns. This function can receive a list of columns and would only read these.
Another question, I am asking me now, is if the row group is the correct abstraction, or if we go even smaller there is also iter_batches https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetFile.html#pyarrow.parquet.ParquetFile.iter_batches. However, this would also read a row group, but if a conversion from parquet to arrow is needed, it would avoid allocating memory for the whole row group.

pyexasol/callback.py
```diff
@@ -74,33 +78,88 @@ def import_from_pandas(pipe, src, **kwargs):
...
+            for field in row_group_table.schema:
+                if types.is_nested(field.type):
+                    raise ValueError(
+                        f"Field {field} of schema from file {file} is hierarchical which is not supported."
+                    )
+
+            write_options = csv.WriteOptions(include_header=False, **kwargs)
+            csv.write_csv(row_group_table, pipe, write_options=write_options)
```
tkilias, line 162: Also the order in which we write the columns, might be something people want to customize.

pyexasol/callback.py
ArBridgeman: Since the initial specifications were for local files, I updated to memory_map=True & using iter_batches.

I switched where **kwargs was passed, which was WriteOptions, to instead go to the iter_batches. Now, columns can be selected & in their desired output order.

pyexasol/callback.py
ArBridgeman: Is it ok for the order of the columns to come from the iter_batches?

pyexasol/callback.py
```diff
@@ -74,33 +78,88 @@ def import_from_pandas(pipe, src, **kwargs):
...
+        csv,
+        parquet,
+        types,
+    )
+
+    if not (parquet_files := get_parquet_files(source)):
+        raise ValueError(f"source {source} does not match any files")
+
+    for file in parquet_files:
+        parquet_file = parquet.ParquetFile(file)
```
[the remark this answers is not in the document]
ArBridgeman: URLs & remote sources would be outside the scope of this issue.

pyexasol/callback.py
```diff
@@ -74,33 +78,99 @@ def import_from_pandas(pipe, src, **kwargs):
...
+
+    Args:
+        source: Local filepath specification(s) to process. Can be one of:
+            - list[pathlib.Path]: list of specific files
+            - pathlib.Path: can be either a file or directory. If it is a directory,
+            all files matching the following pattern *.parquet will be processed.
+            - str: representing a filepath which already contains a glob pattern
+            (e.g., "/local_dir/*.parquet")
+        **kwargs:
+            Custom params for "parquet.Table.iter_batches". This can be used
```
tkilias, line 136:
```suggestion
            Custom params for "pyarrow.parquet.Table.iter_batches". This can be used
```

tkilias (approved):

PR #216, opened 2025-08-07
Bugfix/183 fix exahttpthread

### LINKED ISSUE #183: 🐞 'ExaHttpThread' object has no attribute 'terminate_export'
# Summary
In the class `ExaHTTPTransportWrapper` in the function `export_to_callback` in the except block the function `terminate_export` of an object of type `ExaHttpThread `is called. But this function does not exist. 

## Reproducing the Issue

Reproducibility: always

This bug occured in a framework that incorporated PyExasol to load data from multiple tables from Exasol. In case there is an error in the connection (for e.g. a network issue), this bug can be reproduced.

Due to its nature (there must be an error in the data transfer or network) it's a bit difficult to reproduce the bug stand alone. I didn't prepare a MWE. 

## Expected Behaviour
The correct function (`terminate`) to terminate a ExaHttpThread is called. 

## Actual Behaviour
The function `terminate_export` (and `terminate_import`) are called but do not exist. 

## Root Cause (optional)
Function call to `terminate_export` is implemented here: [http_transport.py#L374)](https://github.com/exasol/pyexasol/blob/647afdc46d2a8c3644aeb75b2e066c20b2cab531/pyexasol/http_transport.py#L374)

However, this function does not exist. The correct function to be called is called `terminate`, defined here: [http_transport.py#L292](https://github.com/exasol/pyexasol/blob/647afdc46d2a8c3644aeb75b2e066c20b2cab531/pyexasol/http_transport.py#L292)

Probably, the same holds for the function call `terminate_import`, implemented here: [http_transport.py#L413](https://github.com/exasol/pyexasol/blob/647afdc46d2a8c3644aeb75b2e066c20b2cab531/pyexasol/http_transport.py#L413)

## Context
The function `terminate_export` was added with this commit: [Improve exception handling and pipe close order for HTTP transport th… · exasol/pyexasol@94114ec · GitHub](https://github.com/exasol/pyexasol/commit/94114ec94670a10e914a6d3ea39ff31c85b9bf2a)

And then removed with a following commit: [Use context manager for pipes while calling HTTP transport callback, … · exasol/pyexasol@e12c36f · GitHub](https://github.com/exasol/pyexasol/commit/e12c36f21fa354ab765d6c72eba1e266d04d24bb#diff-e295ced294a62631f7dc9c6e4bf84dc50978f18540c76073439a1c536058de1b)

In this commit, the two functions `terminate_export` and `terminate_import` were combined to `terminate`. But the function call was not changed accordingly.

### System
- PyExasol version 0.26.0

ArBridgeman:
closes #183

### DIFF
--- a/test/unit/http_transport_test.py
+++ b/test/unit/http_transport_test.py
@@ -419,19 +419,6 @@ class TestExaHTTPTransportWrapper:
             )
 
     @staticmethod
-    def test_export_to_callback_fails_at_callback(
-        http_transport_wrapper_with_mocks, mock_http_thread
-    ):
-        mock_http_thread.read_pipe = None
-        with pytest.raises(Exception, match=ERROR_MESSAGE):
-            http_transport_wrapper_with_mocks.export_to_callback(
-                callback=export_callback, dst=None
-            )
-
-        mock_http_thread.terminate.assert_called()
-        mock_http_thread.join.assert_called_once()
-
-    @staticmethod
     def test_import_to_callback_fails_as_not_a_callback(
         http_transport_wrapper_with_mocks,
     ):
@@ -439,16 +426,3 @@ class TestExaHTTPTransportWrapper:
             http_transport_wrapper_with_mocks.import_from_callback(
                 callback="string", src=None
             )
-
-    @staticmethod
-    def test_import_from_callback_fails_at_callback(
-        http_transport_wrapper_with_mocks, mock_http_thread
-    ):
-        mock_http_thread.write_pipe = None
-        with pytest.raises(Exception, match=ERROR_MESSAGE):
-            http_transport_wrapper_with_mocks.import_from_callback(
-                callback=import_callback, src=None
-            )
-
-        mock_http_thread.terminate.assert_called_once()
-        mock_http_thread.join.assert_called_once()


test/unit/http_transport_test.py
```diff
@@ -377,3 +382,73 @@ def test_delimit_raises_exception(export_sql_query):
...
+    def test_export_to_callback_fails_as_not_a_callback(
+        http_transport_wrapper_with_mocks,
+    ):
+        with pytest.raises(ValueError, match="is not callable"):
+            http_transport_wrapper_with_mocks.export_to_callback(
+                callback="string", dst=None
+            )
+
+    @staticmethod
+    def test_export_to_callback_fails_at_callback(
```
ArBridgeman, line 422: as we have an integration test for this case, would we also want to keep this test using mocks?

tomuben: Good question. I would say if the integration is significantly slower than the unit test, it's worth to keep both. Otherwise one of them can be removed.

ArBridgeman: Thanks, the new integrations tests take ~2s total to run, so I'll remove the overlapping unit tests.

tomuben (approved):

PR #220, opened 2025-08-15
Add performance testing start

ArBridgeman:
This is a very rough draft PR. pytest-benchmark can only work in a set number of ways, which are different from tools I've used or made. Throughout the code, there are many thoughts, which are not yet organized into what should be tackled first.

# Checklist

## For any Pull Request

Is the following correct:
* [ ] the title of the Pull Request?
* [ ] the title of the corresponding issue?
* [ ] there are no other open [Pull Requests](../../../../pulls) for the same update/change? 
* [ ] that the issue which this Pull Request fixes ("Fixes...") is mentioned?

## When Changes Were Made

Did you:
* [ ] update the changelog?
* [ ] update the implementation?
* [ ] check coverage and add tests: unit tests and, if relevant, integration tests?

## When Preparing a Release

Have you:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,6 +75,7 @@ exasol-integration-test-docker-environment = "^4.0.0"
 faker = "^24.14.1"
 exasol-toolbox = "^1.8.0"
 types-ujson = "^5.10.0.20250326"
+pytest-benchmark = {extras = ["histogram"], version = "^5.1.0"}
 
 [tool.black]
 line-length = 88
@@ -107,8 +108,7 @@ module = [
     "turbodbc.*",
     # code we are not yet type checking
     "examples.*",
-    "test.unit.*",
-    "test.integration.*",
+    "test.*",
 ]
 ignore_errors = true
 ignore_missing_imports = true
--- a/test/integration/performance/connection_http_transport_benchmark_test.py
+++ b/test/integration/performance/connection_http_transport_benchmark_test.py
@@ -1,0 +1,264 @@
+from abc import (
+    ABC,
+    abstractmethod,
+)
+from inspect import cleandoc
+from pathlib import Path
+from typing import Optional
+
+import pytest
+
+from pyexasol import ExaConnection
+
+"""pytest-benchmark
+- only allows one usage of measurement per test run...
+- SHOULD not include the preparation step, only the sending step
+- can control how many round, etc. with pedantic
+- by default will re-run multiple times without resetting fully -> this means for import
+ that the data becomes much greater over time
+
+need to prepare data in efficient/universal-way as input -> faker?, cheats with export?
+worthwhile with import to create large file or just string together x rows, y times.
+
+do test against
+- csv -> direct DB call
+- file
+- iterable
+- pandas
+- parquet
+- polars
+
+discuss results with Torsten before too much development to see if right direction
+"""
+
+"""
+Hmm, based on an AI response to the question, we have
+Column sizes:
+  SALES_ID            :   18 bytes
+  SALES_TIMESTAMP     :   26 bytes
+  PRICE               :   11 bytes
+  CUSTOMER_NAME       :  200 bytes
+
+CSV formatting overhead:
+  quotes              :   10 bytes
+  delimiters          :    4 bytes
+  newline             :    2 bytes
+
+Total row size: 271 bytes
+
+Rows per GB: ~4 million
+
+Batch processing test:
+Time to generate 100,000 rows in batches: 1.34 seconds
+Estimated time for 4M rows with batch processing: 50.77 seconds (0.01 hours)
+"""
+
+EXPORT_FROM_TABLE = "SALES"
+IMPORT_INTO_TABLE = "TMP_SALES_COPY"
+INITIAL_SIZE = 1_000
+FINAL_SIZE = 4_000_000
+
+
+def create_empty_table(connection: ExaConnection, table_name: str):
+    ddl = cleandoc(
+        f"""
+        CREATE OR REPLACE TABLE {table_name} (
+            SALES_ID                DECIMAL(18,0) IDENTITY NOT NULL PRIMARY KEY,
+            SALES_TIMESTAMP         TIMESTAMP,
+            PRICE                   DECIMAL(9,2),
+            CUSTOMER_NAME           VARCHAR(200)
+          );
+        """
+    )
+    connection.execute(ddl)
+    connection.commit()
+
+
+@pytest.fixture(scope="session")
+def session_connection(connection_factory):
+    con = connection_factory()
+    yield con
+    con.close()
+
+
+@pytest.fixture(scope="session")
+def tmp_source_directory(tmp_path_factory) -> Path:
+    return tmp_path_factory.mktemp("benchmark")
+
+
+@pytest.fixture(scope="session")
+def create_empty_sales_table(session_connection: ExaConnection):
+    create_empty_table(connection=session_connection, table_name=EXPORT_FROM_TABLE)
+    yield EXPORT_FROM_TABLE
+
+    ddl = f"DROP TABLE IF EXISTS {EXPORT_FROM_TABLE};"
+    session_connection.execute(ddl)
+    session_connection.commit()
+
+
+@pytest.fixture(scope="session")
+def create_initial_sales_data(
+    create_empty_sales_table, session_connection: ExaConnection
+):
+    query = f"""
+        INSERT INTO {EXPORT_FROM_TABLE} (SALES_TIMESTAMP, PRICE, CUSTOMER_NAME)
+        SELECT ADD_SECONDS(TIMESTAMP'2024-01-01 00:00:00', FLOOR(RANDOM() * 365 * 24 * 60 * 60)),
+        RANDOM(1, 125),
+        SUBSTRING(
+        LPAD(TO_CHAR(FLOOR(RANDOM() * 1000000)), 6, '0'),
+        1,
+        FLOOR(RANDOM() * 3) + 5)
+        FROM dual
+        CONNECT BY level <= {INITIAL_SIZE}
+    """
+    session_connection.execute(query)
+    session_connection.commit()
+
+
+# def fill_table(connection: ExaConnection):
+#     create_empty_table(connection=connection)
+#     create_initial_data(connection=connection)
+#     # then do table operations to get to final size
+#
+#
+# fill_table(connection=connection)
+class ImportBenchmark(ABC):
+    """
+    Interface class used for benchmarking various import methods.
+    It's recommended in the pytest-benchmark tool to check the output
+    to see that the call finished as expected. Many of the import
+    functions, however, do not have a non-None result.
+    """
+
+    def __init__(self, connection: ExaConnection, source_directory: Path):
+        self.connection = connection
+        self.source_directory = source_directory
+        self.export_from_table_name = EXPORT_FROM_TABLE
+        self.import_into_table_name = IMPORT_INTO_TABLE
+
+    @property
+    @abstractmethod
+    def source_file(self):
+        pass
+
+    @property
+    def source_filepath(self) -> Optional[Path]:
+        return self.source_directory / self.source_file
+
+    def create_empty_table(self):
+        create_empty_table(
+            connection=self.connection, table_name=self.import_into_table_name
+        )
+
+    @abstractmethod
+    def prepare_data(self):
+        pass
+
+    @abstractmethod
+    def execute_import_command(self):
+        pass
+
+    def perform_import(self):
+        self.execute_import_command()
+        count_query = f"SELECT count(*) FROM {self.import_into_table_name};"
+        return self.connection.execute(count_query).fetchval()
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -75,6 +75,7 @@ exasol-integration-test-docker-environment = "^4.0.0"
 faker = "^24.14.1"
 exasol-toolbox = "^1.8.0"
 types-ujson = "^5.10.0.20250326"
+pytest-benchmark = {extras = ["histogram"], version = "^5.1.0"}
```
ArBridgeman, line 78: was trying out, but we don't necessarily need with histogram

test/integration/performance/connection_http_transport_benchmark_test.py
```diff
@@ -0,0 +1,264 @@
...
+        1,
+        FLOOR(RANDOM() * 3) + 5)
+        FROM dual
+        CONNECT BY level <= {INITIAL_SIZE}
+    """
+    session_connection.execute(query)
+    session_connection.commit()
+
+
+# def fill_table(connection: ExaConnection):
```
ArBridgeman, line 118: could use merge/union to double, but how does this benefit us for the import testing? for sure in the export one this is good.

test/integration/performance/connection_http_transport_benchmark_test.py
```diff
@@ -0,0 +1,264 @@
+from abc import (
```
ArBridgeman, line 1: where would we want benchmarking stored later? put into the integration tests for existing fixture usage.

test/integration/performance/connection_http_transport_benchmark_test.py
```diff
@@ -0,0 +1,264 @@
...
+    tmp_source_directory: Path,
+    create_initial_sales_data,
+):
+    # currently, connection runs with NO_CERT & WITH_CERT, for local testing only
+    # NO_CERT was used as WITH_CERT adds the complexity of a VM. considering
+    # our main concern is the performance of the import functions, we might want to
+    # only consider NO_CERT.
+
+    def setup():
+        # Reset state here; need at 0 so data doesn't increase throughout runs
```
ArBridgeman, line 246: not sure how important this is to reset table each time.
when this was initially added, i was using "SALES" for everything.
now that we have "SALES", and "TMP_SALES_COPY", we could, assuming no DB changes, continue to add on to TMP_SALES_COPY -> only thing is it is useful to verify that the code works as desired by checking the output rows (which is we keep adding on to TMP_SALES_COPY makes is a function of iterations.

test/integration/performance/connection_http_transport_benchmark_test.py
tkilias: Yes, this is more for export.

test/integration/performance/connection_http_transport_benchmark_test.py
ArBridgeman: chose to leave as is as not affecting and otherwise we'd have to use pytest-benchmark setup. as within our single iteration, the rounds do not  reset function-scoped fixtures.

.github/workflows/performance-checks.yml
```diff
@@ -0,0 +1,36 @@
...
+
+jobs:
+  performance-checks:
+    name: Run Performance-Checks (Python-${{ matrix.python-version }}, Exasol-${{ matrix.exasol-version}})
+    runs-on:  ubuntu-24.04
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.11"]
```
ArBridgeman, line 14: double-check: we only want for 1 python version & db?

noxfile.py
```diff
@@ -76,3 +104,38 @@ def run_examples(session: Session) -> None:
...
+    """
+    test_path = _ROOT / "test/performance"
+    benchmark_path = test_path / ".benchmarks"
+
+    command = [
+        "pytest",
+        str(test_path),
+        "--benchmark-sort=name",
+        "--benchmark-compare=0001",
+        # the greatest stddev ~1s, so we set this at 1.25 of that for a robust estimate
```
ArBridgeman, line 134: double-check: see this comment & below, but what kind of criteria would be beneficial here? should we split the import & exports or leave as is?

.github/workflows/performance-checks.yml
```diff
@@ -0,0 +1,36 @@
...
+          extras: "all"
+
+      - name: Start test db
+        run: >
+          poetry run -- nox -s db:start -- --db-version ${{ matrix.exasol-version }}
+
+      - name: Run performance tests and compare to saved benchmark
+        run: poetry run -- nox -s test:performance
+
+      - name: Stop test db
```
ArBridgeman, line 35: will see how it goes with the CI run, but we could also save out this file & upload it as an artifact.

noxfile.py
```diff
@@ -76,3 +113,95 @@ def run_examples(session: Session) -> None:
...
+                f"- {test} is not present in {current_benchmark.filepath.name}"
+            )
+            continue
+        elif expected_results is None:
+            errors.append(
+                f"- {test} is not present in {expected_benchmark.filepath.name}"
+            )
+            continue
+
+        if (
```
ArBridgeman, line 195: As the tests will primarily be maintained by the CI, it makes sense that our reference benchmark json should come from CI runs & not our local machines. To facilitate this now & as the values change over time, I've uploaded the result from this run as an artifact.

In order to successfully add an artifact, we need for the result of the nox session `performance:test` to be a zero exit. This means that we cannot use the `--benchmark-compare-fail` in the command, unless we want to do some fun bash stuff to later have it pop out.

Thus, to do the check, we needed to create something like `peformance:check`. It seemed like the py.test-benchmark static check did not include performing a comparison with `--benchmark-compare-fail`, so this is why we ended up with this custom code. This allows us per test to have a specific result in mind without making some not-so-nice assumptions. If you look at the runtimes, we could say something like if it deviates by a 1s as being significant., but per test, the significance of this 1s depends on its stddev. A similar argument can be made for using the %. Thus, I chose here to look at the mean & stddev for guidance, which can be done a per test basis.

noxfile.py
```diff
@@ -76,3 +113,95 @@ def run_examples(session: Session) -> None:
...
+            > expected_results["mean"] + expected_results["stddev"]
+            or current_results["mean"]
+            < expected_results["mean"] - expected_results["stddev"]
+        ):
+            errors.append(
+                f"- mean of {test} is not in {expected_results['mean']} +/- {expected_results['stddev']}"
+            )
+
+    if errors:
+        errors = ["the comparison failed due to:"] + errors
```
ArBridgeman, line 206:
```suggestion
        errors = ["\nthe comparison failed due to:"] + errors
```

noxfile.py
ArBridgeman: (personally, i'd prefer to use the median, but we only have stddev and while we can convert that to the MAD, idk, feel like i'm already pushing the limits here)

ArBridgeman: so the question is: 
- how do we feel about this comparison check? would we want to alter the factor on the stddev from 1 to something else?

.github/workflows/performance-checks.yml
tkilias: Yeah, I think for the moment, one of each is enough. However, we usually work with the even numbered python versions. So, python 3.10 or 3.12. We are switching to 3.12 soon in the UDFs so probably this

noxfile.py
tkilias: Yeah a fixed percentage seems problematic for tests with external systems (DB). Usually we would need something like

https://en.wikipedia.org/wiki/Wilcoxon_signed-rank_test

See also https://or.stackexchange.com/questions/2558/statistical-tests-for-benchmark-comparison

.github/workflows/performance-checks.yml
```diff
@@ -0,0 +1,46 @@
...
+        uses: actions/upload-artifact@v4.6.2
+        with:
+          name: performance-python${{ matrix.python-version }}-exasol${{ matrix.exasol-version}}
+          path: test/performance/.benchmarks/Linux-CPython-3.11-64bit/0002_benchmark_performance.json
+          include-hidden-files: true
+
+      - name: Stop test db
+        run: poetry run -- nox -s db:stop
+
+      - name: Run performance check against saved reference
```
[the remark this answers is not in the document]
tkilias: Agreed, we should describe the process in the developer guide

noxfile.py
tkilias: Only if we can run the benchmarks in the default runner and use a matrix in the CI to run each test in parallel on different runners

tkilias: FYI 

> --benchmark-json=PATH
> Dump a JSON report into PATH. Note that this will include the complete data (all the timings, not just the stats).

https://pytest-benchmark.readthedocs.io/en/latest/usage.html

noxfile.py
```diff
@@ -76,3 +121,105 @@ def run_examples(session: Session) -> None:
...
+    pytest-benchmark as it should more robustly detect when an individual test's
+    performance changes. The pytest-benchmark options as of 0.5.1 focus on overall
+    performance deviations, like +/- 0.1 s or +/- 10% of the mean, which has the
+    underlying assumption that all tests have similar runtime distributions.
+    """
+    # This threshold is used to determine current_benchmark differs from
+    # previous_benchmark. 0.05 is taken for the evaluative criterion of the
+    # pvalue based on common statistical standards when rejecting the null hypothesi
+    # that the runtime data is drawn from the same distribution
+    alpha_threshold = 0.05
```
ArBridgeman, line 183: need to fine tune. at least at n=10 runs at 4M, this is not stable.
dunno if we want secondary criteria, but will discuss steps for moving forward

test/performance/.benchmarks/0001_performance.json
```diff
@@ -0,0 +1,667 @@
...
+                    2.0785514680000006,
+                    2.053984645000014,
+                    2.073814807999952,
+                    2.037034587999983,
+                    1.9895156129999805,
+                    2.0482951599999524,
+                    1.3755199900000434,
+                    2.322247650999998,
+                    2.1706722630000286,
+                    7.116984646999981,
```
tkilias, line 438: The imports outlier's don't look random, mostly the round before the last is slow. Maybe we force a flush to disk of the uncommitted data at this point, because we don't cleanup.

test/performance/.benchmarks/0001_performance.json
```diff
@@ -0,0 +1,667 @@
...
+                    2.624776814000029,
+                    2.6672313569999915,
+                    2.649661206000019,
+                    2.6911374989999786,
+                    2.6888602770000034,
+                    2.6985145380000404,
+                    2.009181237000007,
+                    2.81157267399999,
+                    2.7745194739999306,
+                    7.796403220000002,
```
tkilias, line 603: Again

test/performance/.benchmarks/0001_performance.json
```diff
@@ -0,0 +1,667 @@
...
+                    2.3423490500000526,
+                    2.344218029999979,
+                    2.3032589780000308,
+                    2.3802630760000056,
+                    2.2827221370000643,
+                    2.362977909000051,
+                    1.8534778030000325,
+                    2.4616331400000036,
+                    2.534256818000017,
+                    7.390800105000039,
```
tkilias, line 658: Again

test/performance/connection/http_transport_import_test.py
```diff
@@ -0,0 +1,106 @@
...
+    data = request.getfixturevalue(data_creator)
+
+    def func_to_be_measured():
+        return getattr(connection, import_method)(
+            data,
+            table=empty_import_into_table,
+            import_params={"columns": ["SALES_TIMESTAMP", "PRICE", "CUSTOMER_NAME"]},
+        )
+
+    benchmark.pedantic(func_to_be_measured, iterations=1, rounds=benchmark_specs.rounds)
```
tkilias, line 102: I saw a setup or teardown parameter 

setup (callable) –

A function to call right before calling the target function in the first iteration of every round.

teardown (callable) –

A function to call after the last iteration of every round.

https://share.google/HYkd8TJRkLTfIVe8H

Because we use iterations 1 and multiple rounds, we should be able to cleanup for each measurement

.github/workflows/performance-checks.yml
```diff
@@ -0,0 +1,47 @@
...
+  performance-checks:
+    name: Performance-Tests (Python-${{ matrix.python-version }}, Exasol-${{ matrix.exasol-version}}), Run-${{ matrix.run-number }}
+    runs-on:  ubuntu-24.04
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.12"]
+        exasol-version: ["8.32.0"]
+        run-number: ["1", "2"]
```
tkilias, line 16:
```suggestion
        run-number: ["1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11", "12", "13", "14", "15"]
```

ArBridgeman:
closed as done in #253

PR #242, opened 2025-09-05
Switch more explicitly to test if None; do not make truthy assumptions

### LINKED ISSUE #241: Make check for columns parameter more robust
The conditional `if not self.columns` in this line

https://github.com/exasol/pyexasol/blame/6a10f581ff4ea408563deda9a27ec5a647460d8b/pyexasol/http_transport.py#L115

has replaced a check `if "columns" not in self.params`.

This can cause some problems in cases where the columns parameter is not a list, but list-like. One particular instance happens when you try to import a dataframe into a table with a subset of the table's columns. The following used to work:

    import_from_pandas(df, table, schema, import_params={'columns' : df.columns})

but now it needs to be updated to

    import_from_pandas(df, table, schema, import_params={'columns' : list(df.columns)})

which might not be entirely expected. 

This could be made better by checking if self.columns is None, or alternatively to cast the columns to a list when the SqlQuery is instantiated.

ArBridgeman:
closes #241 

# Checklist

## For any Pull Request

Is the following correct:

## When Changes Were Made

Did you:
* [x] update the changelog?
* [x] resolve any failing CI criteria (incl. Sonar quality gate)?

## When Preparing a Release

Have you:

### DIFF
--- a/pyexasol/http_transport.py
+++ b/pyexasol/http_transport.py
@@ -40,17 +40,19 @@ class SqlQuery:
     row_separator: Optional[str] = None
 
     def _build_csv_cols(self) -> str:
-        if self.csv_cols is None:
-            return ""
+        if self.csv_cols is not None:
+            safe_csv_cols_regexp = re.compile(
+                r"^(\d+|\d+\.\.\d+)(\sFORMAT='[^'\n]+')?$", re.IGNORECASE
+            )
+            for c in self.csv_cols:
+                if not safe_csv_cols_regexp.match(c):
+                    raise ValueError(f"Value [{c}] is not a safe csv_cols part")
 
-        safe_csv_cols_regexp = re.compile(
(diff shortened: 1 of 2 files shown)


pyexasol/http_transport.py
```diff
@@ -30,16 +31,16 @@ class SqlQuery:
...
-    csv_cols: Optional[list[str]] = None
+    csv_cols: Optional[Iterable[str]] = None
     encoding: Optional[str] = None
     format: Optional[str] = None
     null: Optional[str] = None
     row_separator: Optional[str] = None
 
     def _build_csv_cols(self) -> str:
-        if not self.csv_cols:
+        if self.csv_cols is None:
```
MarleneKress79789, line 43: What happens if the iterables where set, but are empty? i think this would return false. is that what you want?

MarleneKress79789 (approved):

PR #252, opened 2025-09-17
Benchmark with export to parquet

ArBridgeman:
Do NOT MERGE. For demo purposes only.


test/performance/connection/http_transport_export_test.py
```diff
@@ -0,0 +1,232 @@
...
+        iterations=1,
+        rounds=benchmark_specs.rounds,
+        warmup_rounds=benchmark_specs.warm_up_rounds,
+    )
+
+    number_rows = file_length_function(export_dst)
+    assert number_rows == benchmark_specs.final_data_size
+
+
+def test_export_to_parquet(
```
ArBridgeman, line 99: A few different variants here; at least on the performance side, there weren't significant deviations in times. We'd likely only retain this test for the export_to_parquet.

ArBridgeman:
initial testing completed. this can be closed & the follow-up done in another pr

PR #352, opened 2026-07-16
#348: Added `ExaConnection.execute_sql_script()` for executing multi-statement SQL scripts

### LINKED ISSUE #348: Add execute_sql_script function
# Summary
We see that we need, in more and more projects, a function like execute_sql_script, which gets an SQL string with multiple statements.

# Details
This is challenging because usually you would need an SQL parser that can parse the SQL dialect. SQLGlot can do this partially, but it doesn't support all constructs of Exasol.
Another option is to write a parser that doesn't parse the whole SQL, but only the statement end markers. This was done in exapump.

## Background & Context
* For exasol/ansible-collection, we are requested to write an exasol_script task (https://github.com/exasol/ansible-collection/issues/79)
* We should generate for each extension a SQL script for installing this extension, and we need to test it after generation

## Examples
```python
execute_sql_script("""
SELECT 1; SELECT 2;
ALTER SESSION SET script_languages=PYTHON=builtin_python;
-- my comment

CREATE OR REPLACE PYTHON3 ADAPTER SCRIPT schema.name() AS
  x = 1;
  y = 2;
/

""")
```

## References
* https://github.com/exasol-labs/exapump/blob/main/src/commands/sql.rs
* https://github.com/exasol-labs/exapump/blob/main/specs/sql/query-output/spec.md
* https://github.com/exasol-labs/exapump/blob/main/specs/sql/execution-flow/spec.md
* https://github.com/exasol/ansible-collection/issues/16
* https://github.com/exasol/transformers-extension/pull/392
* Python translation from exapump's splitter implementation [sql_splitter.py](https://github.com/user-attachments/files/29328856/sql_splitter.py)

kaklakariada:
# Checklist

Fixes #348 

## For any Pull Request

Is the following correct:

## When Changes Were Made

Did you:
* [x] update the changelog?
* [x] update the User Guide & other documentation?
* [x] resolve any failing CI criteria (incl. Sonar quality gate)?

## When Preparing a Release

Have you:
* [x] thought about version number (major, minor, patch)? -> minor
* [x] checked Exasol packages for updates and resolved open vulnerabilities, if easily possible?


ahsimb:
Why not use the [sqlparse](https://pypi.org/project/sqlparse/)?

ahsimb:
Standalone / inside a script body is not quote-aware (pyexasol/_sql_splitter.py, consume_script_body). Once inside SCRIPT_BODY state, the scanner does not track string literals — so a Lua/Python long string containing a line consisting only of / would falsely terminate the script body early. This matches how Exasol's own SQL clients (EXAplus, etc.) parse scripts too, so it's a pre-existing convention rather than a novel bug, but it's worth calling out explicitly in the "Limitations" section of the new doc page, since it's non-obvious and will silently produce a malformed second "statement" rather than an error.

So, if a Python script happens to have a code like
```
a = (b
/
c)
```
The scan will terminate on "/"

ahsimb:
Partial results are discarded on failure. execute_sql_script (connection.py:336-354) is a plain list comprehension: return [self.execute(statement) for statement in split_sql_script(script)]. If statement 3 of 5 fails, the ExaStatement objects for statements 1–2 (already executed against the server) are dropped along with the raised exception — the caller has no way to inspect what succeeded before the failure short of re-running with try/except around a manual loop. The docstring documents the failure semantics but not this loss of prior results; worth a mention, or consider raising a custom exception that carries the completed statements.

ahsimb (dismissed):

ahsimb (approved):

PR #355, opened 2026-07-28
Expand Exasol test version matrix

### LINKED ISSUE #273: Set up PyExasol to run against 2025.x and 7.x
# Summary
The `BaseConfig` provided by the PTB provides the latest DB versions available in the ITDE. In order to run against these, PyExasol should be checked against them and modifications made in the tests. Currently, we do not run against 2025.x or 7.x.

Due to a temporary conflict between Ubuntu OS + Exasol 7.x driver + UDFs, we had replaced 7.x in our tests with 8.x. When the issue has been resolved, we should re-add 7.x into our slow tests. It otherwise works for all other tests beyond those in examples.

tkilias:
closes #273 

## Summary

- retain Exasol 8.29.13 coverage and add 2025.1.11 and 2026.1.0
- make 2026.1.0 the default database version for local Nox database starts
- apply the established 8.x metadata expectations to all database versions 8 and newer

## Why

The integration-test fixtures only handled database major versions 7 and 8. On 2025.1.11 this produced `None` expected values for metadata assertions, causing the failures observed in the linked Actions run.

## Validation

- `python3 -m compileall -q noxconfig.py test/integration/conftest.py`
- `git diff --check`

Full Nox integration tests were not run locally because the Poetry environment did not have the `nox` executable installed.

### DIFF
--- a/noxconfig.py
+++ b/noxconfig.py
@@ -86,5 +86,5 @@ PROJECT_CONFIG = Config(
     # & a known issue in the ITDE & will be resolved in:
     #   https://github.com/exasol/pyexasol/issues/285
     python_versions=("3.10", "3.11", "3.12", "3.13"),
-    exasol_versions=("8.29.13", "2025.1.11", "2026.1.0"),
+    exasol_versions=("8.29.13", "2025.1.11", DEFAULT_DB_VERSION),
 )
--- a/noxconfig.py
+++ b/noxconfig.py
@@ -10,7 +10,7 @@ from pydantic import (
 )
 
 DEFAULT_PORT = 8563
-DEFAULT_DB_VERSION = "8.29.13"
+DEFAULT_DB_VERSION = "2026.1.0"
 CONTAINER_SUFFIX = "test"
 CONTAINER_NAME = f"db_container_{CONTAINER_SUFFIX}"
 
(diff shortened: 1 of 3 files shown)


noxconfig.py
```diff
@@ -86,7 +86,5 @@ def source_code_path(self) -> Path:
     # & a known issue in the ITDE & will be resolved in:
     #   https://github.com/exasol/pyexasol/issues/285
     python_versions=("3.10", "3.11", "3.12", "3.13"),
-    # Changes for 2025.1.x have not yet been made. This will be resolved in:
-    #   https://github.com/exasol/pyexasol/issues/273
-    exasol_versions=("8.29.13",),
+    exasol_versions=("8.29.13", "2025.1.11", "2026.1.0"),
```
ArBridgeman:
```suggestion
    exasol_versions=("8.29.13", "2025.1.11", DEFAULT_DB_VERSION),
```

noxconfig.py
```diff
@@ -10,7 +10,7 @@
 )
 
 DEFAULT_PORT = 8563
-DEFAULT_DB_VERSION = "8.29.13"
+DEFAULT_DB_VERSION = "2026.1.0"
```
ArBridgeman, line 13: Add entry to `unreleased.md` 😄

test/integration/conftest.py
```diff
@@ -315,7 +315,7 @@ def expected_user_table_column_last_visit_ts(db_major_version):
 
     if db_major_version == 7:
         return timestamp_type(size=8, sql_type="TIMESTAMP")
-    elif db_major_version == 8:
+    elif db_major_version >= 8:
```
ArBridgeman, line 320: This looks good; looks like the remaining failing tests are just due to keywords 😊 and can be adapted based on the test output

ArBridgeman (approved):

PR #359, opened 2026-07-30
fix: reject DSN with a fingerprint placed after the port

### LINKED ISSUE #237: DSN parser does not detect invalid patterns
### Checklist
In order to speedup fixing the bug, please make sure you address as many items from the following checklist as possible.

- [x] I have reproduced the issue with the latest released version of PyExasol.
- [ ] I have added references to issues that sound similar in section _Related Issues_.

# Summary
DSN Parser does not detect wrong DSN values, for example "localhost:8563/123" (fingerprint behind port)

## Reproducing the Issue

See unit test `dsn_test.py::test_invalid_fingerprint`.

### Steps to reproduce the behavior:
Create EXAConnection() with dsn = "localhost:8563/123" => pyexasol will try to establish a connection with invalid host "localhost:8563" (note: this is the host name!!!).

## Expected Behaviour

`pyexasol` should detect invalid patterns for DSN correctly and raise an exception **before** trying to establish a connection.

## Actual Behaviour

`pexasol` tries to establish a websocket connection to an invalid host name if the DSN has incorrect format. This might lead to confusing error messages.

## Root Cause (optional)

Regular expression in [connection.py](https://github.com/exasol/pyexasol/blob/6a10f581ff4ea408563deda9a27ec5a647460d8b/pyexasol/connection.py#L1421) are not enough.

### Additional Context (optional)

Possibly the implementation needs to use a real parser instead of just regular expressions, like:
https://github.com/lark-parser/lark

Sanjays2402:
Fixes #237

The DSN hostname group was `.+?`, which also matches `:` and `/`, so an invalid DSN such as `localhost:8563/1234` (fingerprint placed after the port) parsed `localhost:8563` as the *hostname* and pyexasol then tried to connect to that bogus host. Restricting the hostname prefix/suffix groups to `[^:/]` makes the parser raise `ExaConnectionDsnError` before any connection attempt; all valid DSN forms are unchanged.

This change was prepared with AI assistance; the regression test was run locally and fails without the fix.

# Checklist

## For any Pull Request

Is the following correct:

## When Changes Were Made

Did you:
* [x] update the changelog?

## When Preparing a Release

Have you:


test/unit/connection/process_dsn_test.py
```diff
@@ -0,0 +1,52 @@
...
+        assert excinfo.value.message == expected
+
+    @staticmethod
+    @pytest.mark.parametrize(
+        "dsn,expected",
+        [
+            ("localhost", [("localhost", 8563, None)]),
+            ("localhost:8564", [("localhost", 8564, None)]),
+            ("localhost/ABC", [("localhost", 8563, "ABC")]),
+            ("localhost/ABC:8564", [("localhost", 8564, "ABC")]),
```
tomuben, line 38: Maybe add some more test-cases here, to ensure the new regular expression is not too restrictive:

```suggestion
            ("localhost", [("localhost", 8563, None)]),
            ("localhost:8564", [("localhost", 8564, None)]),
            ("localhost/ABC", [("localhost", 8563, "ABC")]),
            ("localhost/ABC:8564", [("localhost", 8564, "ABC")]),
            ("127.0.0.1..2/ABC:8564", [("127.0.0.1..2", 8564, "ABC")]),
            ("my-host-1/ABC:8564", [("my-host-1", 8564, "ABC")]),
            ("my-host-1.com/ABC:8564", [("my-host-1.com", 8564, "ABC")]),
```

Sanjays2402: Added those, with two notes.

The range case expands rather than staying literal — `127.0.0.1..2/ABC:8564` yields two hosts (`127.0.0.1` and `127.0.0.2`), matching the existing `127.0.0.1..2/CDE:8565` case above, so I used that expectation instead of the single tuple.

The `my-host-1*` cases needed `resolve_hostnames=False` on the fixture, otherwise they hit real DNS and fail with `Could not resolve IP address`. Existing cases pass unchanged with that flag.

`test/unit/connection` is 41 passed. Pushed as 63ad6f6.

tomuben: @Sanjays2402 thank you!

tomuben (dismissed):

test/unit/connection/process_dsn_test.py
Sanjays2402: Merged master to clear the conflict. 2.3.1 was cut in a12276d which emptied `unreleased.md`, so my `#237` bugfix entry moved back under a fresh `## Bugfixes` heading there (it is not in `changes_2.3.1.md`). No code changes; `test/unit/connection` is 41 passed on the merged head bfb2b60.

tomuben (approved):

MarleneKress79789:
We cannot merge this PR directly, because it is a fork, and our reporting requires the pr to be from a branch of the repository. I created a duplicate pr and will merge that one and then close this pr. Thank you for your contribution!

ArBridgeman:
Thanks again! This was merged in #406

ArBridgeman:
Thanks @Sanjays2402! This was merged & released in PyExasol 2.4.1:
https://github.com/exasol/pyexasol/releases/tag/2.4.1

PR #368, opened 2026-08-12
Add threading event to coordinate end

### LINKED ISSUE #349: `import_from_iterable` hangs when the license of the DB is expired
# Summary

`import_from_iterable` hangs when the license of the DB is expired. The SQL query returns `Statement currently not allowed because: ...`, but this can only be seen in the log because the http thread import_from_iterable is blocked, and we join it first. It doesn't get released when the SQL query fails.
https://github.com/exasol/pyexasol/blob/75c9055c0870f4686bfe85c9ff1e2dcb38fa527a/pyexasol/connection.py#L987
We probably need to introduce an event that the two threads can signal when they are ready or encounter an error and wait on this event.
https://docs.python.org/3/library/threading.html#threading.Event

ArBridgeman:
closes #349 

Add threading event to coordinate end

When tried without a preceding `.join()`, it would hang indefinitely and would not reach the `.set()`.

Unit & integration test needed to be updated as mock was with join_with_exc, which is no longer used.

Add new integration test to mock DB license error.

# Checklist

## For any Pull Request

Is the following correct:

## When Changes Were Made

Did you:
* [x] update the changelog?
* [ ] update the User Guide & other documentation? -> need to do

## When Preparing a Release

Have you:

### DIFF
--- a/doc/developer_guide/import_export_threading.rst
+++ b/doc/developer_guide/import_export_threading.rst
@@ -43,7 +43,7 @@ are workers created by :meth:`ExaConnection.import_from_callback` or
 HTTP thread internals
 ---------------------
 
-The HTTP thread owns and runs the a TCP Server (``ExaTCPServer``). The server uses an
+The HTTP thread owns and runs a TCP Server (``ExaTCPServer``). The server uses an
 HTTP request handler ('`ExaHttpRequestHandler``) to move data between Exasol and the pipe.
 
 .. mermaid::
--- a/doc/developer_guide/import_export_threading.rst
+++ b/doc/developer_guide/import_export_threading.rst
@@ -40,11 +40,11 @@ are workers created by :meth:`ExaConnection.import_from_callback` or
         sql --> |"executes IMPORT or EXPORT"| db
         sql -.-> |"uses address and terminates on failure"| http
 
-HTTP thread internals
+HTTP Thread Internals
 ---------------------
 
 The HTTP thread owns and runs a TCP Server (``ExaTCPServer``). The server uses an
-HTTP request handler ('`ExaHttpRequestHandler``) to move data between Exasol and the pipe.
+HTTP request handler (``ExaHttpRequestHandler``) to move data between Exasol and the pipe.
 
 .. mermaid::
 
@@ -75,7 +75,7 @@ The pipe separates the callback from the network transport. This lets the
 callback work with a file-like binary stream without needing to know about the
 socket or the HTTP request.
 
-Data flow
+Data Flow
 ---------
 
 The direction through the HTTP thread depends on the operation:
@@ -92,7 +92,7 @@ thread. The coordinator waits for the workers, joins the HTTP thread first, and
 then joins the SQL thread so that a SQL failure can stop the HTTP worker before
 cleanup waits for it.
 
-Understanding failures
+Understanding Failures
 ----------------------
 
 An exception can originate in the callback, the HTTP transport, or the SQL
--- a/doc/developer_guide/import_export_threading.rst
+++ b/doc/developer_guide/import_export_threading.rst
@@ -1,0 +1,113 @@
+Import and Export Threading
+===========================
+
+The :doc:`user guide's import and export overview
+<../user_guide/exploring_features/import_and_export/index>` describes the
+public behavior and links to further usage details. This page complements it
+with an implementation-oriented view of the threads and transport components
+behind callback-based ``IMPORT`` and ``EXPORT`` operations. The diagrams make
+the ownership, data flow, and error-propagation relationships easier to trace
+when developing or debugging PyExasol.
+
+The callback variants of ``IMPORT`` and ``EXPORT`` use three cooperating threads.
+The calling thread runs the user callback, while an HTTP thread transfers data
+and a SQL thread executes the database statement. The HTTP thread and SQL thread
+are workers created by :meth:`ExaConnection.import_from_callback` or
+:meth:`ExaConnection.export_to_callback`.
+
+.. mermaid::
+
+    flowchart TD
+        subgraph callback_group["Callback"]
+            callback["Calling thread"]
+            pipe["OS pipe"]
+
+            callback <--> |"reads or writes data"| pipe
+        end
+
+        subgraph http_group["HTTP thread"]
+            http["ExaHttpThread"]
+        end
+
+        subgraph sql_group["SQL thread"]
+            sql["ExaSQLImportThread or<br/>ExaSQLExportThread"]
+        end
+
+        db[("Exasol database")]
+
+        pipe <--> http
+        http <--> |"HTTP(S) transport"| db
+        sql --> |"executes IMPORT or EXPORT"| db
+        sql -.-> |"uses address and terminates on failure"| http
+
+HTTP thread internals
+---------------------
+
+The HTTP thread owns and runs the a TCP Server (``ExaTCPServer``). The server uses an
+HTTP request handler ('`ExaHttpRequestHandler``) to move data between Exasol and the pipe.
+
+.. mermaid::
+
+    flowchart TD
+        subgraph http_group["HTTP thread"]
+            http["ExaHttpThread"] --> |"owns and runs"| server["ExaTCPServer"]
+            server --> handler["ExaHttpRequestHandler"]
+        end
+
+        handler <--> pipe["OS pipe"]
+        handler <--> |"HTTP(S) request"| db[("Exasol database")]
+
+Responsibilities
+----------------
+
+* **Calling thread:** Starts the workers and runs the user callback. For an
+  ``IMPORT``, the callback writes data to the pipe. For an ``EXPORT``, it reads
+  data from the pipe.
+* **HTTP thread:** Owns a TCP server (``ExaTCPServer``) and runs its request loop. The
+  server passes the request to a request handler (``ExaHttpRequestHandler``), which moves
+  the streamed data through the pipe.
+* **SQL thread:** Builds and executes the ``IMPORT`` or ``EXPORT`` statement. The
+  statement contains the address exposed by the TCP server, so Exasol can
+  connect to the HTTP thread. The SQL thread also keeps a reference to the HTTP
+  thread and terminates it when the SQL operation fails.
+
+The pipe separates the callback from the network transport. This lets the
+callback work with a file-like binary stream without needing to know about the
+socket or the HTTP request.
+
+Data flow
+---------
+
+The direction through the HTTP thread depends on the operation:
+
+* For ``EXPORT``, Exasol sends an HTTP ``PUT`` request to the TCP server.
+  The request handler writes the received data to the pipe, and the callback
+  reads that data from the pipe.
+* For ``IMPORT``, Exasol sends an HTTP ``GET`` request to the TCP server.
+  The callback writes data to the pipe, while the request handler reads it and
+  sends it to Exasol.
+
+In both cases, the SQL thread runs concurrently with the callback and the HTTP
+thread. The coordinator waits for the workers, joins the HTTP thread first, and
+then joins the SQL thread so that a SQL failure can stop the HTTP worker before
+cleanup waits for it.
+
+Understanding failures
+----------------------
+
+An exception can originate in the callback, the HTTP transport, or the SQL
+operation. Since the components share a data path and depend on one another,
+one failure can cause a secondary failure elsewhere. The callback API collects
+the exceptions observed during cleanup in ``ExaImportError`` or
+``ExaExportError``.
+
+For example, if ``ExaTCPServer.handle_request()`` raises a transport exception,
+``ExaHttpThread.run()`` stores it in ``http_thread.exc``. The SQL operation may
+also fail because its ``IMPORT`` or ``EXPORT`` transport was interrupted. Both
+failures can therefore appear in the aggregated exception. Conversely, if the
+SQL thread fails first, it terminates the HTTP thread; the HTTP thread may then
+exit without an independent exception.
+
+For the public behavior and callback variants, see :doc:`the user guide
+<../user_guide/exploring_features/import_and_export/index>` and
+:doc:`HTTP Transport <../user_guide/exploring_features/import_and_export/http_transport>`.
--- a/pyexasol/connection.py
+++ b/pyexasol/connection.py
@@ -995,7 +995,13 @@ class ExaConnection:
             self.options["encryption"],
             thread_event=thread_event,
         )
-        sql_thread = ExaSQLImportThread(self, compression, table, import_params)
+        sql_thread = ExaSQLImportThread(
+            self,
+            compression,
+            table,
+            import_params,
+            thread_event=thread_event,
+        )
 
         try:
             http_thread.start()
@@ -1006,10 +1012,17 @@ class ExaConnection:
             with http_thread.write_pipe as pipe:
                 result = callback(pipe, src, **callback_params)
 
+            # The callback has finished writing. Allow the HTTP handler to send the
+            # final chunk and finish the request.
+            http_thread.server.can_finish_get.set()
+
+            # Either worker may finish first. Wake up as soon as one reports
+            # completion so errors can be observed without waiting on the
+            # other connection indefinitely.
             thread_event.wait()
 
-            http_thread.raise_with_exception()
-            sql_thread.raise_with_exception()
+            http_thread.join_with_exc()
+            sql_thread.join_with_exc()
 
             return result
 
--- a/pyexasol/connection.py
+++ b/pyexasol/connection.py
@@ -987,11 +987,13 @@ class ExaConnection:
             False if ("format" in import_params) else self.options["compression"]
         )
 
+        thread_event = threading.Event()
         http_thread = ExaHttpThread(
             self.ws_ipaddr,  # type: ignore
             self.ws_port,  # type: ignore
             compression,
             self.options["encryption"],
+            thread_event=thread_event,
         )
         sql_thread = ExaSQLImportThread(self, compression, table, import_params)
 
@@ -1004,8 +1006,13 @@ class ExaConnection:
             with http_thread.write_pipe as pipe:
                 result = callback(pipe, src, **callback_params)
 
-            http_thread.join_with_exc()
-            sql_thread.join_with_exc()
+            http_thread.join()
+            sql_thread.join()
+
+            thread_event.wait()
+
+            http_thread.raise_with_exception()
+            sql_thread.raise_with_exception()
 
             return result
 
--- a/pyexasol/connection.py
+++ b/pyexasol/connection.py
@@ -1006,9 +1006,6 @@ class ExaConnection:
             with http_thread.write_pipe as pipe:
                 result = callback(pipe, src, **callback_params)
 
-            http_thread.join()
-            sql_thread.join()
-
             thread_event.wait()
 
             http_thread.raise_with_exception()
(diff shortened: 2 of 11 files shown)


ArBridgeman:
@codex

pyexasol/connection.py
```diff
@@ -1004,8 +1006,13 @@ def import_from_callback(
...
 
-            http_thread.join_with_exc()
-            sql_thread.join_with_exc()
+            http_thread.join()
+            sql_thread.join()
+
+            thread_event.wait()
+
+            http_thread.raise_with_exception()
+            sql_thread.raise_with_exception()
```
ArBridgeman: To do:

- [ ] Check documentation to see if changes needed
- [ ] Do similar change to `export_to_callback`

ArBridgeman:
@codex

pyexasol/connection.py
```diff
@@ -1004,6 +1012,15 @@ def import_from_callback(
             with http_thread.write_pipe as pipe:
                 result = callback(pipe, src, **callback_params)
 
+            # The callback has finished writing. Allow the HTTP handler to send the
+            # final chunk and finish the request.
+            http_thread.server.can_finish_get.set()
```
tkilias, line 1035: what I am asking me now, does callback(pipe,...) ever return when the sql query fails, but the http connection is still healthy

tkilias: ok the SQLThread is calling terminate on the HTTPThread that hopefull should do it

tkilias: Which in fact is calling self.server.can_finish_get.set(). Slowly I am annoyed by this event, that it is so cluttered. Could you create a list of places where it is set?

ArBridgeman: In this PR, the `set()` is called in 3 places:

* (added here in connection.py) https://github.com/exasol/pyexasol/blob/bug/349_add_thread_event/pyexasol/connection.py#L1017
* http_thread.join() (and by extension http_thread.join_with_exc) https://github.com/exasol/pyexasol/blob/bug/349_add_thread_event/pyexasol/http_transport.py#L470
* http_thread.terminate() https://github.com/exasol/pyexasol/blob/bug/349_add_thread_event/pyexasol/http_transport.py#L481

ArBridgeman: `can_finish_get.wait()` is used in 2 places:
* https://github.com/exasol/pyexasol/blob/bug/349_add_thread_event/pyexasol/http_transport.py#L893
* https://github.com/exasol/pyexasol/blob/bug/349_add_thread_event/pyexasol/http_transport.py#L917

test/integration/import_and_export/export_to_callback_test.py
```diff
@@ -99,18 +107,159 @@ def test_custom_export_callback(
 @pytest.mark.exceptions
 class TestExportToCallbackExceptions:
     @staticmethod
-    def test_export_callback_has_exception(connection, empty_table):
+    def test_export_callback_has_exception(
+        connection, empty_table, capture_callback_threads
+    ):
         error = ValueError("Error from callback")
 
         def export_cb(pipe, dst, **kwargs):
```
ckunki: Could we rename this function to `raise_error()`? 
I hope this benefits the readability of the test setup, like
```python 
connection.export_to_callback(callback=raise_error, ...)
```

pyexasol/connection.py
```diff
@@ -906,14 +906,20 @@ def export_to_callback(
             False if ("format" in export_params) else self.options["compression"]
         )
 
+        thread_event = threading.Event()
```
ckunki: Sorry for focusing on the naming once again.

Could we propose a name for `thread_event` with a bit more semantics, e.g. 
* `has_terminated`
* `success`
* `failure`
* `thread_exception`
* `thread_terminated`
* `thread_finished`
* `finish_event`

Or is it required to be very generic here? 

I assume the event signals that either one of the two threads involved here, has terminated either successfully or with an error. And that in consequence the other thread should be terminated, too, right?

If this is correct, then I would hope to find as few words as possible, expressing this?

Alternatively, we could also create a subclass of `threading.Event()` and add a more expressive docstring?

pyexasol/connection.py
```diff
@@ -925,6 +931,15 @@ def export_to_callback(
...
+            # The callback has finished writing. Allow the HTTP handler to send the
+            # final chunk and finish the request.
+            http_thread.server.can_finish_get.set()
+
+            # Either worker may finish first. Wake up as soon as one reports
+            # completion so errors can be observed without waiting on the
+            # other connection indefinitely.
+            thread_event.wait()
+
             http_thread.join_with_exc()
```
ckunki, line 944: If I understand correctly, then there is also an asymmetry here:
* The `ExaSQLThread` can terminate the `ExaHttpThread` (and maybe even *should* do so!)
* But not the other way round

If I understand correctly, then this also imposes the ordering of the join commands: `http_thread` must come first!

If I am correct, then this constraint should be expressed a bit more clearly. At least by some comments, but better by programming constructs preventing a misuse.

test/integration/import_and_export/export_to_callback_test.py
```diff
@@ -99,18 +107,159 @@ def test_custom_export_callback(
...
+                    callback=export_cb, dst=None, query_or_table=empty_table
+                )
+
+        assert len(ex.value.exceptions) == 1
+        assert ex.value.exceptions[0] == error
+        assert not http_thread.is_alive()
+        assert not sql_thread.is_alive()
+
+    @staticmethod
+    def test_closed_ws_connection(
```
ckunki, line 139: Why does this test need a default fixture `connection` when it creates a new connection anyway?

test/integration/import_and_export/export_to_callback_test.py
```diff
@@ -99,18 +107,159 @@ def test_custom_export_callback(
...
+        with capture_callback_threads(ExaSQLExportThread) as (
+            http_thread,
+            sql_thread,
+        ):
+            with pytest.raises(ExaExportError) as ex:
+                new_connection.export_to_callback(
+                    export_cb_with_close, None, empty_table
+                )
+
+        assert len(ex.value.exceptions) == 2
```
ckunki, line 167: Can you explain, why we expect 2 exceptions to be raised?
Also, I hope a short explanation about the *causality* to be really helpful.
E.g. 
1. test case simulates web socket to be closed 
2. in turn `ExaSQLThread` terminates and also terminates the `ExaHttpThread`
3. hence the test case expects both thread to not be alive anymore.

test/integration/import_and_export/export_to_callback_test.py
```diff
@@ -99,18 +107,159 @@ def test_custom_export_callback(
...
+        connection,
+        output_filepath,
+        fill_table,
+        export_cb,
+        capture_callback_threads,
+    ):
+        error = BrokenPipeError("Broken pipe in http_thread")
+
+        def write_final_chunk_with_exception(_handler):
+            raise error
```
ckunki: Variable `error` seems to be not used anywhere else: 

```suggestion
        def write_final_chunk_with_exception(_handler):
            raise BrokenPipeError("Broken pipe in http_thread")
```

ckunki (changes requested):

test/integration/import_and_export/export_to_callback_test.py
ArBridgeman: Good catch. That's a copy and paste error from the original one I had in `import_from_callback_test.py`. I've removed both 😄

pyexasol/connection.py
ArBridgeman: Always happy to have naming convention discussions :smile:, especially here, as I truly struggled to come up with one that fit.

Because of its nature, where the `thread_even`t changes states throughout it's lifecycle and its intended usage, I was thinking of calling it  `worker_finished_event`.

Yes, you are correct that `worker_finished_event` is a shared notification event. Either worker (sql_thread or http_thread) sets it when it finishes, successfully or exceptionally.

The event only wakes the coordinator. It does not itself terminate the other worker:

  - SQL failure terminates the HTTP worker so callback I/O is interrupted.
  - HTTP failure does not terminate SQL directly; coordinator cleanup handles that.
  - Successful completion of one worker does not immediately terminate the other.

Thus, to me, the name `worker_finished_event` reflects this “at least one worker finished” behavior.

What do you think?

pyexasol/connection.py
ArBridgeman: Yes, you are correct that there is an asymmetry here and that the join order is intentional.

Yes, `http_thread` must be joined before `sql_thread` because the SQL worker can terminate the HTTP worker, while the HTTP worker does not terminate SQL directly. Joining HTTP first ensures that an SQL  failure has already had an opportunity to stop it before the coordinator waits for it.

And also, I agree that this should be expressed more clearly. I'll add a comment for it.

As I'm adding quite a bit of comments, please let me know if you have other suggestions for how to document it. Your expertise is this area is always appreciated :green_heart: .

test/integration/import_and_export/export_to_callback_test.py
ArBridgeman: good catch; will simplify the code

test/integration/import_and_export/import_from_callback_test.py
```diff
@@ -136,16 +140,30 @@ def test_custom_import_callback(
 @pytest.mark.exceptions
 class TestImportFromCallbackExceptions:
     @staticmethod
-    def test_import_callback_has_exception(connection, empty_table):
+    def test_import_callback_has_exception(
+        connection, empty_table, capture_callback_threads
+    ):
+        """
+        The defined callback function raises an exception:
```
ckunki, line 147: Sorry, I needed quite some time to understand the test scenario and intention. 


Image

Assuming my understanding (based on which I created the image) to be correct, I would propose the following wording: 

* The callback function is assumed to write data to the HTTP thread's pipe.
* Simulate the callback function to raise an error.
* The error causes the pipe to close.
* The SQL thread in turn cannot complete the data IMPORT as the incoming data stream is interrupted.
* The SQL statement IMPORT used by the SQL thread hence aborts as incomplete and raises an `ExaQueryError`.
* The HTTP thread does not recognize any error but is only stopped during cleaning of the whole import operation.

ckunki: With the former docstring I had some problems: 

> The defined callback function raises an exception:
>   - That failure closes the write pipe before the IMPORT request can finish.

Why is it a "write pipe"? Is there also a "read pipe"? Probably not in this scenario.
* But then the word used in the docstring is either over-specific or confusing
* Also the test's implementation only contains 1 reference to the pipe which is
  not used in our simulated callback

>   - The SQL thread therefore receives an incomplete request from Exasol and
>     raises an ``ExaQueryError``.

Why does the SQL thread receive an incomplete request?  How does Exasol know
about the HTTP thread, it's pipe, and a failed callback?

>   - The HTTP thread is only stopped by cleanup; it does not perform the failed
>   SQL operation, and it,therefore, is not expected to have its own exception.

I assumed the SQL operations are performed by the SQL thread, not by the HTTP
thread. Or am I wrong, here? Maybe the term "it" is confusing or means
different things in different places?

In general, I think, some terms and phrases are based on and inspired by the
actual implementation but leave the abstraction visible in the test case.

ckunki: But once again, following up on our discussion and the shared feelings, I think the problems in findings the right words and terms is also caused by an implementation with a bit unclear structure, repetition, and also in some places using imprecise words and even concepts itself.

test/integration/import_and_export/import_from_callback_test.py
```diff
@@ -154,38 +172,131 @@ def import_cb(pipe, src, **kwargs):
...
+        empty_table,
+        import_cb,
+        all_data,
+        capture_callback_threads,
     ):
-        with patch("pyexasol.connection.ExaHttpThread.join_with_exc") as mock:
-            mock.side_effect = BrokenPipeError("Broken pipe in http_thread")
+        """
+        The HTTP thread raises an exception while writing a chunk:
+         - The handler catches that failure and closes its transport.
```
ckunki:
```suggestion
        The RequestHandler used by the HTTP thread's TCPServer 
        raises an exception while writing a chunk of data.
        - The HTTP thread catches this exception and closes its transport
```

test/integration/import_and_export/import_from_callback_test.py
```diff
@@ -154,38 +172,131 @@ def import_cb(pipe, src, **kwargs):
...
+        connection, input_filepath, empty_table, import_cb, capture_callback_threads
+    ):
+        """
+        ``ExaTCPServer.handle_request`` raises an exception in the HTTP
+        thread:
+          - Because the HTTP server can no longer carry the data, the
+            concurrently running SQL thread's IMPORT receives a transport
+            failure and raises an ``ExaQueryError``.
+          - The callback only supplies data and completes normally.
+        """
```
ckunki: This docstrig is really on the point in my eyes. At least I can understand it without any problems. 
Maybe we just needed some more attempts until having found the right words?

Instead of "HTTP server" I would prefer "TCP server" to stick to the identical words as in "ExaTCPServer" and to avoid confusion with "HTTP thread".

test/integration/import_and_export/import_from_callback_test.py
```diff
@@ -194,25 +305,95 @@ def test_http_thread_has_exception(
...
+        import_cb,
+        capture_callback_threads,
+    ):
+        """
+        The SQL thread raises an exception before it can consume the import data:
+          - The SQL thread terminates the HTTP thread and closes the pipe.
+          - The callback then attempts to use that pipe and raises a ``ValueError``.
+          - The HTTP thread is only being terminated by the SQL thread, so it is not
+            expected to report a separate exception.
+        """
```
ckunki, line 329: Very good, I think!

ckunki: Maybe we should really add a figure to the developer guide (or even user guide?).
In order to introduce all the names, concepts, components, words, and their interrelationships. 

Then, it's hopefully easier pointing to places that can cause errors?

ckunki (dismissed):

pyexasol/connection.py
ArBridgeman: Marking as resolved, as during our conversation August-21, we had agreed. Changes in thoughts are still welcome though :smile:

test/integration/import_and_export/export_to_callback_test.py
ArBridgeman: done

test/integration/import_and_export/export_to_callback_test.py
ArBridgeman: removed in both the import and export tests

test/integration/import_and_export/export_to_callback_test.py
ArBridgeman: done

test/integration/import_and_export/import_from_callback_test.py
ArBridgeman: Thanks, I tried to update it

doc/developer_guide/import_export_threading.rst
```diff
@@ -0,0 +1,113 @@
...
+
+        pipe <--> http
+        http <--> |"HTTP(S) transport"| db
+        sql --> |"executes IMPORT or EXPORT"| db
+        sql -.-> |"uses address and terminates on failure"| http
+
+HTTP thread internals
+---------------------
+
+The HTTP thread owns and runs the a TCP Server (``ExaTCPServer``). The server uses an
```
kratz00:
```suggestion
The HTTP thread owns and runs a TCP Server (``ExaTCPServer``). The server uses an
```

test/integration/import_and_export/import_from_callback_test.py
```diff
@@ -154,38 +172,134 @@ def import_cb(pipe, src, **kwargs):
...
+        )
+        assert not sql_thread.is_alive()
+        assert not http_thread.is_alive()
+
+    @staticmethod
+    def test_http_thread_captures_transport_exception(
+        connection, input_filepath, empty_table, import_cb, capture_callback_threads
+    ):
+        """
+        The TCP server (``ExaTCPServer'', owned by the HTTP thread) raises a transport
```
kratz00: ExaTCPServer <- the markup seems to be malformed?

test/integration/import_and_export/import_from_callback_test.py
```diff
@@ -136,16 +140,30 @@ def test_custom_import_callback(
...
+    def test_import_callback_has_exception(
+        connection, empty_table, capture_callback_threads
+    ):
+        """
+        The defined callback function raises an exception:
+          - That failure closes the write pipe before the IMPORT request can finish.
+          - The SQL thread therefore receives an incomplete request from Exasol and
+            raises an ``ExaQueryError``.
+          - The HTTP thread is only stopped by cleanup; it does not perform the failed
+          SQL operation, and it,therefore, is not expected to have its own exception.
```
kratz00:
```suggestion
          SQL operation, and therefore, it is not expected to have its own exception.
```

test/integration/import_and_export/import_from_callback_test.py
```diff
@@ -154,38 +172,134 @@ def import_cb(pipe, src, **kwargs):
             "Following error occured while reading data"
```
kratz00, line 172:
```suggestion
            "Following error occurred while reading data"
```
I know, not part of this PR - maybe it can be fixed nevertheless

pyexasol/connection.py
```diff
@@ -925,6 +933,18 @@ def export_to_callback(
             with http_thread.read_pipe as pipe:
                 result = callback(pipe, dst, **callback_params)
 
+            # The callback has finished writing. Allow the HTTP handler to send the
```
kratz00:
```suggestion
            # The callback has finished. Allow the HTTP handler to send the
```
Depending on export or import - it's either _writing_ or _reading_, or?

pyexasol/connection.py
```diff
@@ -925,6 +933,18 @@ def export_to_callback(
...
+            # final chunk and finish the request.
+            http_thread.server.can_finish_get.set()
```
kratz00: the export uses PUT, what is GET flagged here?

test/integration/import_and_export/export_to_callback_test.py
ArBridgeman: Did my best to explain in the docstring

test/integration/import_and_export/import_from_callback_test.py
ArBridgeman: Sadly not. It doesn't come from PyExasol. It comes from the database:
ETL-5105: Following error occured while reading data from external connection

ArBridgeman: https://docs.exasol.com/db/latest/changelogs/5241.htm

kratz00: Damn :( Thanks for checking it!

pyexasol/connection.py
ArBridgeman: Good catch! I'd initially added it and missed removing it when adding tests.
I've removed it.

doc/developer_guide/import_export_threading.rst
```diff
@@ -0,0 +1,113 @@
...
+        pipe <--> http
+        http <--> |"HTTP(S) transport"| db
+        sql --> |"executes IMPORT or EXPORT"| db
+        sql -.-> |"uses address and terminates on failure"| http
+
+HTTP thread internals
+---------------------
+
+The HTTP thread owns and runs a TCP Server (``ExaTCPServer``). The server uses an
+HTTP request handler ('`ExaHttpRequestHandler``) to move data between Exasol and the pipe.
```
kratz00:
```suggestion
HTTP request handler (``ExaHttpRequestHandler``) to move data between Exasol and the pipe.
```

kratz00 (approved):

PR #392, opened 2026-09-02
Add native parquet import

ArBridgeman:
This is a test setup using generative AI.
Changes would be done in two separate PRs:
1. for the builder (underway)
2. for the ExaConnection implementation

### DIFF
--- a/pyexasol/http_transport.py
+++ b/pyexasol/http_transport.py
@@ -10,6 +10,7 @@ import threading
 import zlib
 from collections.abc import Iterable
 from dataclasses import dataclass
+from pathlib import Path
 from ssl import SSLContext
 from typing import TYPE_CHECKING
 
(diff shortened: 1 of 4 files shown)


test/integration/import_and_export/parquet_test.py
```diff
@@ -89,7 +89,12 @@ def test_load_single_file(
         self, tmp_path, empty_table, connection, table_name, all_data
     ):
         filepath = tmp_path / "single_file.parquet"
-        self._create_parquet_file(filepath, all_data.list_dict)
+        parquet_table = pa.Table.from_pylist(all_data.list_dict)
```
ArBridgeman, line 92: @tkilias This is the test I was running it on.

pyexasol/http_transport.py
```diff
@@ -796,3 +797,176 @@ def write_success_headers(self):
...
+    def __init__(self, *args, parquet_files: list[Path], **kwargs):
+        self.parquet_files = parquet_files
+        super().__init__(*args, **kwargs)
+
+    @property
+    def total_size(self) -> int:
+        return sum(parquet_file.stat().st_size for parquet_file in self.parquet_files)
+
+
+class ExaParquetHttpRequestHandler(socketserver.StreamRequestHandler):
```
tkilias, line 816: uh it implements HTTP completely from scratch, maybe a question  if we can use https://docs.python.org/3/library/http.server.html#http.server.BaseHTTPRequestHandler, butt also not sure if it is wise

